PR #94, opened 2020-10-24
fix steamID url, improve !trades & !trade messages

rocanruca:
-Fixed the steamID in the url for steamProfile, backpackTF and steamREP (this problem is only when using !trade command, it show admin steamID instead of trade partner steamID).
-Added some improvements in messages for !trade & !trades commands.

### DIFF
--- a/src/classes/Commands.ts
+++ b/src/classes/Commands.ts
@@ -2247,7 +2247,8 @@ export = class Commands {
 
             reply += `\n- Offer #${offer.id} from ${
                 offer.data.partner
-            } (reason: ${offer.data.action.meta.uniqueReasons.join(', ')})`;
+            } (reason: ${offer.data.action.meta.uniqueReasons.join(', ')})
+            ⚠️ Send "!trade ${offer.id}" for more details.\n `;
         }
 
         this.bot.sendMessage(steamID, reply);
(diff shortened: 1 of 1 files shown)


src/classes/Commands.ts
```diff
@@ -2247,7 +2247,8 @@ export = class Commands {
 
             reply += `\n- Offer #${offer.id} from ${
                 offer.data.partner
-            } (reason: ${offer.data.action.meta.uniqueReasons.join(', ')})`;
+            } (reason: ${offer.data.action.meta.uniqueReasons.join(', ')})
+            ⚠️ Send "!trade ${offer.id}" for more details.\n `;
```
idinium96, line 2251: please separate this, and why you put `\n` at the end of it?
```
            reply += `\n- Offer #${offer.id} from ${
                offer.data.partner
            } (reason: ${offer.data.action.meta.uniqueReasons.join(', ')})` +
                '⚠️ Send "!trade ${offer.id}" for more details.';
```

src/classes/Commands.ts
```diff
@@ -2324,8 +2325,9 @@ export = class Commands {
                     : ')');
         }
 
-        const links = (this.bot.handler as MyHandler).tradePartnerLinks(steamID.toString());
-        reply += `\n\nSteam: ${links.steamProfile}\nBackpack.tf: ${links.backpackTF}\nSteamREP: ${links.steamREP}`;
+        const links = (this.bot.handler as MyHandler).tradePartnerLinks(offerData.partner.toString());
+        reply += `\n\nSteam: ${links.steamProfile}\nBackpack.tf: ${links.backpackTF}\nSteamREP: ${links.steamREP} 
+        \n⚠️ Send "!accept ${offerId}" to accept or "!decline ${offerId}" to decline this offer.`;
```
idinium96, line 2330: also separate this.

idinium96 (commented):
please separate these lines of code. thank you.

src/classes/Commands.ts
rocanruca: With `\n` at the end
image

Without `\n` at the end
image
I like it more with `\n` at the end

idinium96: oh I see.. alright you can keep it 👍

PR #115, opened 2020-11-09
add custom listing note parameters

idinium96:
- `notebuy` and `notesell` on `!add` or `!update` commands.

### DIFF
--- a/src/classes/Listings.ts
+++ b/src/classes/Listings.ts
@@ -486,7 +486,46 @@ export = class Listings {
 
         let details: string;
 
-        if (entry.sku === '241;6') {
+        if (entry.notebuy && intent === 0) {
+            details = entry.notebuy
+                .replace(/%price%/g, entry[key].toString())
+                .replace(/%name%/g, entry.name)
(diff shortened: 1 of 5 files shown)


src/classes/Listings.ts
```diff
@@ -486,7 +486,46 @@ export = class Listings {
...
+                details.replace(/%keyPrice%/g, 'Key rate: ' + keyPrice + '/key');
+            } else {
+                details.replace(/%keyPrice%/g, '✨');
+            }
+
+            if (entry.sku === '241;6') {
+                details.replace(/%dueling%/g, '(𝗢𝗡𝗟𝗬 𝗪𝗜𝗧𝗛 𝟱x 𝗨𝗦𝗘𝗦) ');
+            } else {
+                details.replace(/%dueling%/g, '✨');
+            }
```
Bonfire, line 508: Is the replacement for noisemakers with 25 uses also required here?

idinium96: Yea... will add it later after merged into development. Thanks.

PR #132, opened 2020-11-13
Fix scrap adjustment affect the global key rate

sumet123:
- Fix scrap adjustment affect the global key rate

### DIFF
--- a/src/classes/Commands.ts
+++ b/src/classes/Commands.ts
@@ -8,7 +8,7 @@ import TradeOfferManager from 'steam-tradeoffer-manager';
 
 import Bot from './Bot';
 import CommandParser from './CommandParser';
-import { Entry, EntryData } from './Pricelist';
+import { Entry, EntryData, PricelistChangedSource } from './Pricelist';
 import Cart from './Cart';
 import AdminCart from './AdminCart';
 import UserCart from './UserCart';
@@ -1295,7 +1295,7 @@ export = class Commands {
         }
 
         this.bot.pricelist
-            .addPrice(params as EntryData, true, params.sku === '5021;6' ? 'manual' : null)
+            .addPrice(params as EntryData, true, PricelistChangedSource.Command)
             .then(entry => {
                 this.bot.sendMessage(steamID, `✅ Added "${entry.name}".`);
             })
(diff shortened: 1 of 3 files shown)


src/classes/Commands.ts
```diff
@@ -1295,7 +1295,7 @@ export = class Commands {
         }
 
         this.bot.pricelist
-            .addPrice(params as EntryData, true)
+            .addPrice(params as EntryData, true, params.sku === '5021;6' ? 'manual' : null)
```
idinium96, line 1298: what if they add keys with autoprice enabled?

src/classes/Commands.ts
```diff
@@ -1653,7 +1653,7 @@ export = class Commands {
         }
 
         this.bot.pricelist
-            .updatePrice(entryData, true)
+            .updatePrice(entryData, true, entryData.sku === '5021;6' ? 'manual' : null)
```
idinium96, line 1656: same as above. update to enable autoprice

idinium96:
I think it's better to replace `src: string | null = null` in the add/update functions to `isScrapAdjustment = false`, then on the `!add` and `!update` command, it will be unspecified (false), and at Autokeys things, just put `this.isEnableScrapAdjustment`, and make correct logics in validate function.

sumet123:
> I think it's better to replace `src: string | null = null` in the add/update functions (Pricelist.ts) to `isScrapAdjustment = false`, then on the `!add` and `!update` command (Commands.ts), it will be unspecified (false), and at Autokeys.ts, just put `this.isEnableScrapAdjustment`, and make correct logics in validate function in Pricelist.ts.

My first thought is there might be other modules that use `updatePrice()` with different logics so using `isScrapAdjustment` boolean is a bit too specific. (with this idea, the `src` should not be `null` anyways) But for now, I agree that using what you purposed will be more logical.

src/classes/Commands.ts
sumet123: Should be fine? As it won't update key rate (check the logic in `Pricelist.ts`) but the `src` could misleading.

sumet123:
Is the new version any better?

idinium96:
looks like it's better. Have you tried it?

PR #137, opened 2020-11-16
Add test framework

joekiller:
The point of this PR is to add a useful set of tests which can be expanded upon as new features arrive. The test framework is [Jest](https://jestjs.io/en/).

Current goals:

  - Enable asserting the result of SteamUser.chatMessage() calls

Features thus far:

  - Executing in CI
  - Current tests are in `src/classes/__tests__/` which is common Jest convention. It makes imports easy for what is being tested generally because you import ../classname.

### DIFF
--- a/src/classes/__tests__/MyHandler.ts
+++ b/src/classes/__tests__/MyHandler.ts
@@ -1,0 +1,16 @@
+import MyHandler from '../MyHandler';
+import Bot from '../Bot';
+import BotManager from '../BotManager';
+import SteamID from 'steamid';
+import SteamUser from 'steam-user';
+
+jest.mock('SteamUser');
+
+test('onMessage', () => {
+    SteamUser.chatMessage.mockResolvedValue();
+    const botManager = new BotManager();
+    const bot = new Bot(botManager);
+    const myHandler = new MyHandler(bot);
+    const steamID = new SteamID('abc123');
+    myHandler.onMessage(steamID, 'hi');
+});
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,4 @@
 local
 dist
 *.json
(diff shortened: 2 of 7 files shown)


idinium96:
Please only create a new branch from the `development` branch and create a pull request to the `development` branch.

Thank you.
https://github.com/idinium96/tf2autobot/blob/master/.github/contributing.md

joekiller:
Damn I thought I rebased on development.

joekiller:
@idinium96 I think the PR just defaulted to master. I swapped it.

src/classes/__tests__/MyHandler.ts
```diff
@@ -0,0 +1,16 @@
+import MyHandler from '../MyHandler';
+import Bot from '../Bot';
+import BotManager from '../BotManager';
+import SteamID from 'steamid';
+import SteamUser from 'steam-user';
+
+jest.mock('SteamUser');
+
+test('onMessage', () => {
+    SteamUser.chatMessage.mockResolvedValue();
```
idinium96, line 10: maybe you should add `const client: SteamUser = new SteamUser()` before this line and replace `SteamUser.chatMessage` with `client.chatMessage` then let see if this check failure disappears.

idinium96:
I think I will not add this.

joekiller:
Hrm I didn't intend this to be ready. To keep track of progress would you prefer I make an issue that the branch will fulfill? At this point the code is supposed to be an example of where the starting point of testing is.

joekiller:
Yes rtfm. https://github.com/idinium96/tf2autobot/blob/master/.github/contributing.md#pull-requests

I'll make an issue called add testing that'll track my progress and then open the pr

PR #158, opened 2020-11-20
Add Full Backpack information, Update getBackpackSlots method

Bonfire:
An example of the Steam chat message sent to the user when either party has a full backpack:
image

And the Discord Webhook message:
image

`getBackpackSlots()` has been refactored and now returns both `used` and `total` backpack slots from backpack.tf


Bonfire:
This branch can be merged into v2 if need be. Original work for #158 was done in `idinium96:issue-134` so I continued off of that branch and created a PR to merge back into that branch

idinium96:
using `getBackpackSlots()` to get the `used` slots seems not very practical, as backpack.tf might not completely load their backpack information by the time the bot trading is with them.

Bonfire:
> using `getBackpackSlots()` to get the `used` slots seems not very practical, as backpack.tf might not completely load their backpack information by the time the bot is trading with them.

Will make changes to use a more reliable source of backpack slot information

idinium96:
is it still working as intended?

idinium96:
I think it's better to rebase this to v2, or v2-test.

Bonfire:
> is it still working as intended?

Yes, the functionality is the same after making the changes. The only issue I could foresee is there being an issue with fetching the trade partner's max slots, resulting in the value being undefined.

idinium96:
Hurmmm... I see.
There are other options:
1. https://steamapi.xpaw.me/#IEconItems_440/GetPlayerItems
2. https://steamcommunity.com/profiles/76561198053911511/inventory/json/440/2?l=english&count=5000

The first one as what I know is an old API, and people in tf2-automatic discord server told me not to use it, idk why, maybe because it always broke (the previous versions, where people always complain about "Failed to obtain backpack slots, retrying in 10 minutes..." warn log.
The second one is the new one, which I believe Zeus shared, but I don't have any clue where to find total backpack slots from there.

Issue #162, opened 2020-11-20
Add confirmation step to !delete command


fredalex6:
**Is your feature request related to a problem? Please describe.**
Currently, when an admin wants to `!remove` an item the person can confuse the `!remove` command with `!delete`. Similar words, but `!delete` can have serious consequences (deleting an item you meant to remove from the pricelist). Therefore, the two must be separated so that they do not get confused with each other. 

**Describe the feature you'd like**
I suggest a confirmation step to the `!delete` command. To ensure that user do not accidentally mistaken `!remove` for `!delete`. The extra step informs the user on what he/she is doing, that way it decreases the chance of mistaking. I propose two types: 

Example 1 (same as `!remove all=true`):
```
User: !delete name=Mann Co. Supply Crate Key
Bot: ⚠️ Are you sure that you want to delete Mann Co. Supply Crate Key? This will completely delete it from my inventory. Try again with i_am_sure=yes_i_am
User: !delete name=Mann Co. Supply Crate Key&i_am_sure=yes_i_am
Bot: ✅ Successfully deleted Mann Co. Supply Crate Key!
```

Example 2:
```
User: !delete name=Mann Co. Supply Crate Key
Bot: ⚠️ Are you sure that you want to delete Mann Co. Supply Crate Key? This will completely delete it from my inventory (Yes/No)
User: Yes
Bot: ✅ Successfully deleted Mann Co. Supply Crate Key!
```

Bonfire:
Updated your issue, correcting `Mann Co. Supply Create Key` to `Mann Co. Supply Crate Key`

PR #163, opened 2020-11-20
Add confirmation to !delete command

Bonfire:
Example output from these changes:
image

These changes address enhancements described in #162


idinium96 (commented):
Please add confirmation at line 861 (when `sku === null`)

Bonfire:
> Please add confirmation at line 861 (when `sku === null`)

Good catch. All set!

PR #176, opened 2020-11-27
V2 type updates

joekiller:
Add more thorough types for steam-client.

This will help with the mock interface down the line.
TradeManager still needs this treatment but that'll have to be later.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1583,6 +1583,16 @@
                 "@types/bluebird": "*"
             }
         },
+        "@types/bytebuffer": {
+            "version": "5.0.42",
+            "resolved": "https://registry.npmjs.org/@types/bytebuffer/-/bytebuffer-5.0.42.tgz",
+            "integrity": "sha512-lEgKojWUAc/MG2t649oZS5AfYFP2xRNPoDuwDBlBMjHXd8MaGPgFgtCXUK7inZdBOygmVf10qxc1Us8GXC96aw==",
+            "dev": true,
+            "requires": {
+                "@types/long": "*",
+                "@types/node": "*"
+            }
+        },
         "@types/caseless": {
             "version": "0.12.2",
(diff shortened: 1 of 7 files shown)


joekiller:
@idinium96 this is ready for review

idinium96:
Thanks joe, but this not includes the `steam-tradeoffer-manager` enums right?
https://github.com/idinium96/tf2autobot/blob/v2-type-updates/src/classes/Trades.ts#L46

joekiller:
> Thanks joe, but this not includes the `steam-tradeoffer-manager` enums right?
> https://github.com/idinium96/tf2autobot/blob/v2-type-updates/src/classes/Trades.ts#L46

Yes that does address it from the perspective of running code. but it leaves the symbol for the compiler at a member  `TradeOfferManager.ETradeOfferState` that is a string vs this value is the value of ` TradeOfferManager.ETradeOfferState.Active`.  

It is very valid but the symbol is slightly different. It's getting a key from a map (what enums eventaully are) vs a property in itself.

The change is not necessary but I'd suggest we do it at some point. No immediately necessary. 

The types currently in here will help my v2-mock-handler

idinium96:
> > Thanks joe, but this not includes the `steam-tradeoffer-manager` enums right?
> > https://github.com/idinium96/tf2autobot/blob/v2-type-updates/src/classes/Trades.ts#L46
> 
> Yes that does address it from the perspective of running code. but it leaves the symbol for the compiler at a member `TradeOfferManager.ETradeOfferState` that is a string vs this value is the value of ` TradeOfferManager.ETradeOfferState.Active`.
> 
> It is very valid but the symbol is slightly different. It's getting a key from a map (what enums eventaully are) vs a property in itself.
> 
> The change is not necessary but I'd suggest we do it at some point. No immediately necessary.
> 
> The types currently in here will help my v2-mock-handler

Oh already, then this should be fine. Final checks before merging...

Issue #180, opened 2020-11-29
!config command crashes 


christianpug:
**Describe the bug**

**To Reproduce**

1. goto your steam chat and goto your bot
2. .type !config manualReview.enable=false
3. it crashes

**Expected behavior**
> 0|Pootis_Bot  | TypeError: Cannot read property 'exceptionValue' of undefined
0|Pootis_Bot  |     at Object.updateOptionsCommand (/root/tf2autobot/dist/classes/Commands/optionsCommands.js:94:46)
0|Pootis_Bot  |     at Commands.processMessage (/root/tf2autobot/dist/classes/Commands/Commands.js:209:30)
0|Pootis_Bot  |     at MyHandler.onMessage (/root/tf2autobot/dist/classes/MyHandler/MyHandler.js:291:23)
0|Pootis_Bot  |     at Bot.onMessage (/root/tf2autobot/dist/classes/Bot.js:543:22)
0|Pootis_Bot  |     at Immediate.<anonymous> (/root/tf2autobot/dist/classes/Bot.js:163:21)
0|Pootis_Bot  |     at processImmediate (node:internal/timers:463:21)
0|Pootis_Bot  | 2020-11-29 16:40:31 error: TF2Autobot crashed! Please create an issue with the following log:
0|Pootis_Bot  | package.version: 2.0.1; node: v15.2.1 linux x64}
0|Pootis_Bot  | Stack trace:
0|Pootis_Bot  | TypeError: Cannot read property 'exceptionValue' of undefined
0|Pootis_Bot  |     at Object.updateOptionsCommand (/root/tf2autobot/dist/classes/Commands/optionsCommands.js:94:46)
0|Pootis_Bot  |     at Commands.processMessage (/root/tf2autobot/dist/classes/Commands/Commands.js:209:30)
0|Pootis_Bot  |     at MyHandler.onMessage (/root/tf2autobot/dist/classes/MyHandler/MyHandler.js:291:23)
0|Pootis_Bot  |     at Bot.onMessage (/root/tf2autobot/dist/classes/Bot.js:543:22)
0|Pootis_Bot  |     at Immediate.<anonymous> (/root/tf2autobot/dist/classes/Bot.js:163:21)
0|Pootis_Bot  |     at processImmediate (node:internal/timers:463:21)

**Screenshots and logs**
![screenshot](https://user-images.githubusercontent.com/37443483/100548120-ae5fbf80-3238-11eb-947e-6aceff118fa2.PNG)
[...]

PR #181, opened 2020-11-30
parse arrays in CommandParser with tests

joekiller:
This should simplify handing of arrays for the parser. Fixing #180


joekiller:
okay looking good now.

PR #182, opened 2020-11-30
Better autobump feedback

joekiller:
This will give the user feedback for in case autorelist never got enabled when they configured autobump.

If the user is premium and enabled autobump it just silently fails because autorelist is set to false by default.

### DIFF
--- a/src/classes/Commands/optionsCommands.ts
+++ b/src/classes/Commands/optionsCommands.ts
@@ -10,6 +10,7 @@ import { JsonOptions, removeCliOptions } from '../Options';
 
 import { deepMerge } from '../../lib/tools/deep-merge';
 import validator from '../../lib/validator';
+import log from '../../lib/logger';
 
 export function optionsCommand(steamID: SteamID, bot: Bot): void {
     const liveOptions = deepMerge({}, bot.options) as JsonOptions;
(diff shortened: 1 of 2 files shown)


idinium96:
what is this for? I don't understand the point.

joekiller:
I added it because it was confusing to me that I didn't get any messages about autobump when I enabled it. The reason for no messages was due to autorelist being false and premium was true.
 
It seems like the intention of https://github.com/idinium96/tf2autobot/pull/182/commits/d844e35c6cd4643405073bc4746077a1b919351a#diff-eb9b17dfed38b4112a892229008ee55ea2d788890a52f8ba69b17458d00aef7cR113 is to inform the user that autorelist is unnecessary however that message would only appear if they had autobump enabled, started the bot without premium and then at some point got premium with autobump enabled and then the bot would say you don't need this feature.



I added this additional message that would inform if you started the bot with autobump enabled but also had premium as I found it confusing why I wasn't being told that I didn't need it.

joekiller:
Updated my comment to point to https://github.com/idinium96/tf2autobot/blob/d844e35c6cd4643405073bc4746077a1b919351a/src/classes/Listings.ts#L113

idinium96:
Ahh, I see. This should be nice to add.
Thanks!

Issue #190, opened 2020-12-01
Fix occasionally exclusion of item amount


fredalex6:
**Describe the bug**
Sometimes when the bot sends a trade summary, it mistakenly excludes the number of items whenever the item amount is greater than 1. The error prevents the user/admin to know the number of items in a trade, which can be central for further actions or assessments.

**To Reproduce**
1. Trade completes (Bot).
2. The bot sends trade summary to a Discord channel(s) (Discord Webhook).
3. Occasionally excludes the number of items whenever the item amount is greater than 1.

**Expected behaviour**
Whenever the bot sends a trade summary, it has to make sure the trade amount value matches. This counts for all appearances of item amount in summaries.

**Screenshots**
![Issue, amount 1_LI](https://user-images.githubusercontent.com/69604246/100763342-54582900-33f5-11eb-939c-5fe29d17c575.jpg)
![Screenshot (111)](https://user-images.githubusercontent.com/69604246/100763732-dd6f6000-33f5-11eb-8f9b-b1a3ed97c455.png)

**Additional context**
The bot only excludes the number of items in some trades. You can also view this error in the trade sharing channels in TF2Autobot, "#🔴｜community" and "#🔴｜community-2".

**Version:** TF2Autobot@v2.1.0

Bonfire:
Will be fixing this issue later tonight

idinium96:
> Will be fixing this issue later tonight

Thanks @Bonfire

PR #192, opened 2020-12-02
Fix item amount summary

Bonfire:
This change should fix issues where item amounts weren't shown when the bot had `"showStockChanges": false`. Addresses #190 

Example output from changes shown below (show stock changes is disabled):
image


Bonfire:
Noticed that these changes include changes made by @idinium96 from the Development branch. Feel free to move commits around as needed.

Bonfire:
Updated to include summaries that don't have a link (Steam summaries). Further testing for these changes can be seen below.

image
image

PR #207, opened 2020-12-03
🔁 update time taken to have numeric precision

Bonfire:
This PR also includes thresholds for DayJS times. No longer will we have to deal with `44 seconds = 1 minute`!


idinium96:
Can you help refactor all that extends from `src/classes/Commands/Commands.ts` to `src/lib/tools/time.ts `?

PR #208, opened 2020-12-03
Add outgoing message history to admin !message command

Bonfire:
These changes add message history to outgoing messages from the admin, just like incoming messages to the admin. Results from implementation can be seen below.

With `discordWebhook.messages.url` set:
image

With `discordWebhook.messages.url` not set, or `discordWebhook.messages.enable = false`:
image

(Note that in the second image you see two messages. This is because I sent the message to myself so I got the message confirmation and the message itself)


src/lib/DiscordWebhook/sendAdminMessage.ts
```diff
@@ -0,0 +1,50 @@
...
+    bot: Bot
+): void {
+    const opt = bot.options.discordWebhook;
+    const botInfo = (bot.handler as MyHandler).getBotInfo();
+
+    /*eslint-disable */
+    const discordAdminMsg: Webhook = {
+        username: opt.displayName ? opt.displayName : botInfo.name,
+        avatar_url: opt.avatarURL ? opt.avatarURL : botInfo.avatarURL,
+        content: `<@!${opt.ownerID}>, message sent! - ${steamID}`,
```
idinium96, line 24: maybe ping not needed here?

src/classes/Commands/messageCommand.ts
```diff
@@ -75,8 +76,29 @@ export default function message(steamID: SteamID, message: string, bot: Bot): vo
...
+
+        // Send a notification to the admin with message contents & details
+        if (opt.discordWebhook.messages.enable && opt.discordWebhook.messages.url !== '') {
+            sendAdminMessage(steamID.toString(), reply, adminDetails, links, time.time, bot);
+
+            // Send a confirmation message to the admin if they use discord webhooks (in case it isn't apparent enough)
+            bot.sendMessage(
+                steamID,
+                '✅ Your message has been sent.' +
+                    '\n🔍 Please view the channel associated with your "discordWebhook.messages.url" for more information'
```
idinium96, line 90: I think this is not needed.

idinium96: the one that tells you to look into that channel.

PR #231, opened 2020-12-12
fix runaway check function

joekiller:
bind(this) always returns a new function reference which is why it wouldn't unbind

https://stackoverflow.com/a/22870717/2240967

### DIFF
--- a/src/classes/Listings/Listings.ts
+++ b/src/classes/Listings/Listings.ts
@@ -27,8 +27,6 @@ export default class Listings {
 
     private autoRelistEnabled = false;
 
-    private isForceDisabled = false;
-
     private autoRelistTimeout;
 
     private get isAutoRelistEnabled(): boolean {
@@ -41,6 +39,8 @@ export default class Listings {
 
     private templates: { buy: string; sell: string };
 
+    private readonly checkFn;
+
     constructor(bot: Bot) {
         this.bot = bot;
         this.templates = {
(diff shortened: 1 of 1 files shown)


src/classes/Listings/Listings.ts
```diff
@@ -57,26 +59,19 @@ export default class Listings {
...
 
     disableAutorelistOption(): void {
-        if (!this.autoRelistEnabled) {
-            // If the process already not running, just don't execute this.
-            return;
-        }
-        this.bot.listingManager.removeListener('heartbeat', this.checkAccountInfo.bind(this));
-        this.bot.listingManager.off('heartbeat', this.checkAccountInfo.bind(this));
-        this.isForceDisabled = true;
-        this.disableAutoRelist();
```
idinium96, line 79: Thanks, but I wonder why you remove this and replace it with the same thing inside this function.

idinium96: I mean, `this.disableAutoRelist();`
https://github.com/TF2Autobot/tf2autobot/blob/58f6ee85c5ffb0d1c7a0d5b71115ba2088666b0a/src/classes/Listings/Listings.ts#L150-L155

joekiller: I think I was just leaving it how it was prior to the attempted fixes. I cleaned it up and put it through disableAutoRelist now

idinium96:
tested and it's working
image

PR #237, opened 2020-12-14
Add startup time to "onReady" summary

Bonfire:
Example of the new startup summary: 
```2020-12-14 00:39:43 info: TF2Autobot v2.2.5 is ready | 1 item in pricelist | Listings cap: 70 | Startup time: 6s```

### DIFF
--- a/src/classes/MyHandler/MyHandler.ts
+++ b/src/classes/MyHandler/MyHandler.ts
@@ -280,7 +280,7 @@ export default class MyHandler extends Handler {
                 true
             )} in pricelist | Listings cap: ${String(
                 this.bot.listingManager.cap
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/classes/MyHandler/MyHandler.ts [hunk 1]
```diff
@@ -280,7 +280,7 @@ export default class MyHandler extends Handler {
                 true
             )} in pricelist | Listings cap: ${String(
                 this.bot.listingManager.cap
-            )} | Startup time: ${process.uptime().toFixed(0)}ms`
+            )} | Startup time: ${process.uptime().toFixed(0)}s`
```
idinium96, line 283: Please add a space between the value and the unit.

PR #246, opened 2020-12-22
ensure getPricesTF returns a price parsed with Currency()

joekiller:
- tighten prices.tf interface
  - fix optimistic type conversions

### DIFF
--- a/src/lib/ptf-api.ts
+++ b/src/lib/ptf-api.ts
@@ -1,33 +1,129 @@
-import { UnknownDictionary } from '../types/common';
 import { OptionsWithUrl, ResponseAsJSON } from 'request';
 
 import request from 'request-retry-dayjs';
 
-export function getSchema(): Promise<UnknownDictionary<any>> {
+export interface PricesResponse {
+    success: boolean;
+    message?: string;
+}
+
+export interface GetSchemaResponse extends PricesResponse {
+    version: string;
+    time: number;
+    raw: any;
+}
+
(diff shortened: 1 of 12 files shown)


src/lib/ptf-api.ts
```diff
@@ -1,33 +1,129 @@
...
     return apiRequest('GET', `/items/${sku}/history`, { src: source });
 }
 
-export function getSales(sku: string, source: string): Promise<UnknownDictionary<any>> {
+export function getSales(sku: string, source: string): Promise<GetItemSalesResponse> {
     return apiRequest('GET', `/items/${sku}/sales`, { src: source });
 }
 
-export function requestCheck(sku: string, source: string): Promise<UnknownDictionary<any>> {
+export function requestCheck(sku: string, source: string): Promise<PricesResponse> {
```
idinium96, line 118: This type might be wrong.

https://github.com/TF2Autobot/tf2autobot/blob/30cc4416c68c2fe435e590efaee4539b2a0ea01c/src/classes/Commands/Commands.ts#L1057

https://github.com/TF2Autobot/tf2autobot/blob/30cc4416c68c2fe435e590efaee4539b2a0ea01c/src/classes/MyHandler/offer/accepted/updateListings.ts#L40

joekiller: thanks I'll fix these

joekiller: I believe it is fixed now

PR #252, opened 2020-12-24
better linting config

joekiller:
🎨 clean up code and tests 🔨

- make MyHandler the handler of Bot
- better linting config
- remove getHandler as it is unnecessary
- add minor tests for pricelist
- add ptf-api mocks
- ensure mocks and tests don't build into prod
- allow failsafe to detect required options in Options object


src/classes/Bot.ts
```diff
@@ -61,7 +60,7 @@ export default class Bot {
 
     readonly tf2gc: TF2GC;
 
-    readonly handler: Handler;
+    readonly handler: MyHandler;
```
idinium96, line 63: Hurm you sure? 🤔

joekiller: We create new MyHandler() in the constructor so it should be safe.

joekiller: See https://github.com/TF2Autobot/tf2autobot/pull/252/files#diff-450b22a2f8e3cb5c78358968ec6eed9abf6f1c6e6719786a39f467777e7e5d83R162

the handler is a MyHandler type, might as well use it as that

idinium96: alrighty

PR #282, opened 2021-01-21
✅ add removeCliOptions and schema validation tests

joekiller:
currently failing due to permissive validation 🐛

once these tests pass then the bug is fixed.


idinium96:
oppss

PR #288, opened 2021-01-24
✨ Add alternative prices url

### LINKED ISSUE #274: Alternative prices.tf URL
As a user I'd like to use an alternative prices.tf URL.

Feature: user can set alternative autoprice source other than prices.tf

joekiller:
Set CUSTOM_PRICER_URL="" to a url other than https://api.prices.tf to use a different price source.
PRICESTF_API_TOKEN renamed to CUSTOM_PRICER_API_TOKEN

Closes #274


template.env
```diff
@@ -14,6 +14,7 @@ GROUPS=["103582791469033930"]
 ALERTS=["trade"]
 
 PRICESTF_API_TOKEN=""
+PRICESTF_URL=""
```
idinium96, line 17: maybe best to just change these two to other name, such as, `CUSTOM_PRICER_URL` and `CUSTOM_PRICER_API_TOKEN`

src/classes/BotManager.ts
```diff
@@ -30,11 +32,12 @@ export default class BotManager {
 
     private exiting = false;
 
-    constructor(pricestfApiToken?: string) {
-        this.schemaManager = new SchemaManager({});
+    constructor(pricestfUrl?: string, pricestfApiToken?: string) {
+        this.priceSource = new PricesTf(pricestfUrl, pricestfApiToken);
+        this.schemaManager = new SchemaManager({ apiKey: pricestfApiToken });
```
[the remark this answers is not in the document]
joekiller: Okay I understand now

template.env
joekiller: Yes. Thanks for the feedback

src/classes/Pricer.ts
```diff
@@ -0,0 +1,138 @@
...
+    source: string;
+    time: number;
+    buy: KeysMetal | null;
+    sell: KeysMetal | null;
+}
+
+export interface KeysMetal {
+    keys: number;
+    metal: number;
+}
```
idinium96, line 79: This can be replaced by `Currencies` by importing it from `tf2-currencies-2` library.

tsconfig.json
```diff
@@ -1,5 +1,6 @@
 {
     "compilerOptions": {
+        "baseUrl": ".",
```
idinium96, line 3: Is this need to be edited by the user? If so, then tsconfig.json need to be included in `.gitignore` file I guess.

idinium96: or people that want to use a custom pricer will need to fork the repo in order to edit this file?

idinium96: well but then, other people will have trouble compiling the code because tsconfig.json file will never be found.

joekiller: no this just says, the project starts from this tsconfig location. it shouldn't need to be editted.

joekiller:
I'm good with this feature for now.

tsconfig.json
idinium96: I see. noted

joekiller:
There is a bug in this still which I need to fix. Occurs after most transactions.

idinium96:
Noted sir. 👌

joekiller:
https://github.com/TF2Autobot/tf2autobot/commit/b967960f6bc0cd46dc38d40fe66669906b73250f fixed the bug I was tracking so I think this is good now. I've been running on this branch with 3.0.1 master merged and 3.0 prior and beta merged prior etc. just did a rebase.

joekiller:
nope still there. I was just hoping b/c it looked like it. I'll chase it down...

idinium96:
image

joekiller:
Okay dokie, I'm testing this on my bots. No errors so far. I pulled out anything in the interface that wasn't actually used so that functionality isn't misleading.

idinium96:
Alrighty, sir!

joekiller:
This seems ready. 🚀

idinium96:
If I merge this, then it should be v3.2.0.. so fast ⚡

idinium96 (approved):
Looks good although never tested it. Maybe good to merge with `development` and test this right away.

PR #302, opened 2021-02-05
v3.0.4 - important patch

idinium96:
- 🐛 fix wrong altered amount on `!buy`/`!sell` commands.

image
image1
image2

### DIFF
--- a/src/classes/Carts/UserCart.ts
+++ b/src/classes/Carts/UserCart.ts
@@ -415,8 +415,8 @@ export default class UserCart extends Cart {
                         'I only have ' +
                         pluralize(this.bot.schema.getName(SKU.fromString(sku), false), ourAssetids.length, true);
 
-                    // Add the max amount to the cart
-                    this.addOurItem(sku, amount);
+                    // Add the max amount to the cart and substract added amount from command (if exist)
+                    this.addOurItem(sku, this.our[sku] ? amount - this.our[sku] : amount);
                 }
             }
 
@@ -434,7 +434,8 @@ export default class UserCart extends Cart {
                         skuCount.mostCanTrade
                     )}`;
 
-                    this.addOurItem(sku, amount);
+                    // Add the amount we can trade and substract added amount from command (if exist)
+                    this.addOurItem(sku, this.our[sku] ? skuCount.mostCanTrade - this.our[sku] : skuCount.mostCanTrade);
                 }
             }
 
@@ -495,8 +496,8 @@ export default class UserCart extends Cart {
                         'you only have ' +
                         pluralize(this.bot.schema.getName(SKU.fromString(sku), false), theirAssetids.length, true);
 
-                    // Add the max amount to the cart
-                    this.addTheirItem(sku, amount);
+                    // Add the max amount to the cart and and substract added amount from command (if exist)
+                    this.addTheirItem(sku, this.their[sku] ? amount - this.their[sku] : amount);
                 }
             }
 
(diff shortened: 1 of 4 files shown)


idinium96:
The cause from the screenshots (line 437):

https://github.com/TF2Autobot/tf2autobot/blob/af218cffa3b8e883dbd0039f076f5c3887c96735/src/classes/Carts/UserCart.ts#L424-L439


Right?

idinium96:
Also the reason for subtracting `this.our[sku]` or `this.their[sku]` because we already add it here right?

https://github.com/TF2Autobot/tf2autobot/blob/af218cffa3b8e883dbd0039f076f5c3887c96735/src/classes/Commands/Commands.ts#L350-L352

src/classes/Carts/UserCart.ts
```diff
@@ -434,7 +434,8 @@ export default class UserCart extends Cart {
                         skuCount.mostCanTrade
                     )}`;
 
-                    this.addOurItem(sku, amount);
+                    // Add the amount we can trade and substract added amount from command (if exist)
+                    this.addOurItem(sku, this.our[sku] ? skuCount.mostCanTrade - this.our[sku] : skuCount.mostCanTrade);
```
joekiller, line 438: if line 432 `"I can only sell ${skuCount.mostCanTrade}"` is correct (which is appears to be) then I think this line needs to be

`this.addOurItem(sku, skuCount.mostCanTrade);` because that would add the most we listed that could trade.

src/classes/Carts/UserCart.ts
```diff
@@ -415,8 +415,8 @@ export default class UserCart extends Cart {
                         'I only have ' +
                         pluralize(this.bot.schema.getName(SKU.fromString(sku), false), ourAssetids.length, true);
 
-                    // Add the max amount to the cart
-                    this.addOurItem(sku, amount);
+                    // Add the max amount to the cart and substract added amount from command (if exist)
+                    this.addOurItem(sku, this.our[sku] ? amount - this.our[sku] : amount);
```
joekiller, line 419: This line shouldn't change because it is always only adding amount which is equal to the total amount of assets we have or the amount requested.

src/classes/Carts/UserCart.ts
```diff
@@ -495,8 +496,8 @@ export default class UserCart extends Cart {
                         'you only have ' +
                         pluralize(this.bot.schema.getName(SKU.fromString(sku), false), theirAssetids.length, true);
 
-                    // Add the max amount to the cart
-                    this.addTheirItem(sku, amount);
+                    // Add the max amount to the cart and and substract added amount from command (if exist)
+                    this.addTheirItem(sku, this.their[sku] ? amount - this.their[sku] : amount);
```
joekiller, line 500: I think this should be reverted

src/classes/Carts/UserCart.ts
```diff
@@ -513,7 +514,11 @@ export default class UserCart extends Cart {
                         skuCount.mostCanTrade
                     )}`;
 
-                    this.addTheirItem(sku, skuCount.mostCanTrade);
+                    // Add the amount we can trade and substract added amount from command (if exist)
+                    this.addTheirItem(
+                        sku,
+                        this.their[sku] ? skuCount.mostCanTrade - this.their[sku] : skuCount.mostCanTrade
+                    );
```
joekiller, line 521: revert back to skuCount.mostCanTrade as I don't see any evidence of trades where the text says "I can't buy more" only "I can only sell more" which makes me think ony line 437 needs to be updated to match line 516.

joekiller (commented):
I think only one line needs to be changed. Please review and argue otherwise.

idinium96:
Experiment:

image

**Case 1: revert everything, change only `amount` to `skuCount.mostCanTrade` in line 437.**
Result:
image

**Case 2: current commit**
Result:
image
image

Conclusion: Yes you're correct. 👍

arik123 (approved):
Everything looks fine

PR #309, opened 2021-02-07
Timed profit

arik123:
Added timed profit capability to profit calculator.
Would be best for someone to add profit last 24h to stats.


idinium96:
What will this actually do?

arik123:
It adds the ability to set start time since which timedProfit will be calculated - which could be used for showing profit in last x hours / days

idinium96:
Ohhh understood. Thank you 👍

src/lib/DiscordWebhook/sendStats.ts
```diff
@@ -11,11 +11,16 @@ export default function sendStats(bot: Bot, forceSend = false, steamID?: SteamID
     const optDW = bot.options.discordWebhook;
     const botInfo = bot.handler.getBotInfo;
     const trades = stats(bot);
-    const profits = profit(bot);
+    const profits = profit(bot, Date.now() - 86400000);
```
idinium96, line 14: This in milliseconds right? Should be divided by 1000 and round the value so it will match with pollData timestamps (Unix, in seconds).

idinium96 (changes requested):

idinium96 (approved):

PR #337, opened 2021-02-13
✨ an option to only update buying price if an item is in stock

### LINKED ISSUE #284: Better Pricing
Yes, TF2AutoBot will sell items for less than the price the item was bought at. Making the whole gaining 0.11 ref + profit logic redundant.

An option for users to choose how often the items in their backpack which they are selling get re-priced. Items which are in the pricelist, but the bot doesn't have (buying), should be updated frequently, so that the most competitive prices are offered. 
For example:
- If stock = 1 or more, bot doesnt update the selling price, only the buying price. 
- If stock = 0, bot updates both the buying and selling prices
  o Safe guard should be added here: if buying price = selling price of other bot users (simple bp.tf check) then buying price - 0.11 ref = new buying price. But the selling price stays same with the market (other bots). 

After 7 days (or whatever the user chooses, can be a entry in options.json?) the whole pricelist stock = 1 or more AND 0 gets updated with new prices. So that the bots inventory turnover remains high. 

Manual pricing but that's yuck. 

**Additional context**
The above allows for profit to be generated in majority if not all of the trades the bot conducts. Not more loss on sales. As prices.tf has cycles in the pricing of its items, by having the above allows for users and their bots to control more of the pricing cycle.

idinium96:
- closes #284
- not implement: _Safe guard should be added here: if buying price = selling price of other bot users (simple bp.tf check) then buying price - 0.11 ref = new buying price. But the selling price stays same with the market (other bots)._

## How does this work?
- Previously, on start, the bot will get full pricelist from prices.tf/custom-autopricer and then loop through the bot pricelist and compare the time. If time in the bot pricelist is less than the time in prices.tf/custom-autopricer pricelist entry, the bot will update the buying and selling prices along with the time.
- This feature, if `pricelist.onlyUpdateBuyingPriceIfInStock.enable` is set to `true` AND that particular item is currently in stock (> 0) AND difference in time is less than threshold (default is 604800 seconds or 7 days, minimum can set is 86400 seconds or 1 day - you can set your own by editing `pricelist.onlyUpdateBuyingPriceIfInStock.thresholdInSeconds` value),
    - then if it's determined that the new buying price (prices.tf/custom-autopricer) is less than the current selling price (in pricelist entry), then update ONLY the buying price of that item.
    - then if it's also determined that the new selling price (prices.tf/custom-autopricer)  is more than the current selling price (in pricelist entry), then update the selling price of that item.
    - if nothing above match, then just ignore updating the prices
- Else if `pricelist.onlyUpdateBuyingPriceIfInStock.enable` is set to `false` AND/OR that particular item is NOT currently in stock (= 0) AND/OR difference in time exceed the threshold, then update all buying/selling/time to the latest one.

- The above conditions also applied when receiving prices.tf/custom-autoprice data (socket).


idinium96:
merging...

Issue #341, opened 2021-02-14
Constructing trade with weapons


Dex-Spirit:
**Describe the bug**
Bot takes trade partner's weapons instead of picking metal using `!buy` / `!sell` command.

**To Reproduce**

1. Request trade with a command
(I don't really know what conditions are needed for this to happen)

**Expected behavior**
Take trade partner's metal or give change using our metal.
Take their weapons ONLY if they don't have enough metal or 0.05 ref needed.

**Screenshots and logs**
![image](https://user-images.githubusercontent.com/61100980/107887423-a85fad80-6f0e-11eb-9b91-d8accdbcd39d.png)

Logs: [Angel_trade_construction_log.txt](https://github.com/TF2Autobot/tf2autobot/files/5978399/Angel_trade_construction_log.txt)
![21;10;13 - 14 02 2021](https://user-images.githubusercontent.com/61100980/107887391-7e0df000-6f0e-11eb-9358-a89e1adfa783.png)

Version: 3.3.0

Issue #281, opened 2021-01-21
Add item trade statistics


Bonfire:
**Is your feature request related to a problem? Please describe.**
Currently, users are unable to determine whether or not specific items are profitable, whether or not specific items buy/sell often, or view general trade data about specific items.

**Describe the feature you'd like**
I propose a new feature - we can call it `!summary`, or perhaps `!sales`, that would provide historic item trade data through Discord Webhook or Steam Chat. It would be called like so: `!summary/sales sku=459;6` or `!summary/sales name=Cosa Nostra Cap`. And it would provide users information like what is seen below:

![image](https://user-images.githubusercontent.com/5704760/105281802-efaa8680-5b7a-11eb-8ef5-88aeedd0486a.png)
_Note: This information was sourced from https://gladiator.tf/_

This would allow users to see if an item buys/sells well. If it has gone up or down over time. How many (quantity) are being bought/sold. Overall, this would provide users with more granular trade data for items, which they can then use to make informed trading decisions.

**Describe alternatives you've considered**
Using another service (like gladiator.tf) to see trade data. However, this is not pertinent to _your specific bot_, which may change things (such as not having an over-/under-cutter, having more/less capital, etc.).

**Additional context**
I believe that this feature could be implemented by either parsing the already existing `polldata.json`, or by logging item-specific trade history in a separate file, such as `history.json`.

idinium96:
Love it, and let's also discuss the best structure for the .json files for this, and I think I like separate files better since the bot will only need to load small files at a time.

PR #343, opened 2021-02-15
Prepare v3.4.0

idinium96:
## Added
- ✨`!itemstats` command - to get statistics of specific items (#281) - @arik123 / @idinium96
    - Added new env variable: `ITEM_STATS_WHITELIST` (default is an empty array, you can put valid SteamIDs here to whitelist people that can use `!itemstats` command on your bot).

## Changed
- 💡 refactor: changed hot rod sheen emoji from `🎗` to `🌸` - @Bonfire
- 🔄 use `@tf2autobot/tradeoffer-manager` library instead of `steam-tradeoffer-manager` - @arik123
    - get rid of the `deepEqual` function, which uses a lot of CPU power.
    - reduce `pollInterval` from 30 seconds to 5 seconds.
- 🔄 increase pricelist count (>4000) & decrease delay time (from 1000 ms to 400 ms) on automatic checking for missing listings (and `!refreshlist` command) - @idinium96
- 🔨 do not auto-add painted items if spelled - @idinium96
- 🔨 do not perform `checkBySKU` for pure and weapons (that is not in the pricelist) - @idinium96
- 🔊 show more activities in debug logs on the `checkBySKU` function - @idinium96
- ✅ solve TODO in `Listings.ts`: Perform check if the item is already in the making - @idinium96
- ✅ determine whether to pick a weapon when constructing an offer (via chat), and 🛑 limit to only pick 1 weapon, and only once (#341) - @idinium96
- ⛔ automatically block banned people (chat) - @idinium96
- 🚫 do not givePrice to crates/cases - @idinium96
- 🔕 Ignore the auto-retry error and remove never read code - @idinium96
- 🎨 use better sentence - decline taking free items - @idinium96


idinium96:
*do not check pure and weapons

Issue #363, opened 2021-02-22
Add support for custom pricers throughout the application


Bonfire:
**Is your feature request related to a problem? Please describe.**
Currently, TF2Autobot displays one of two sources for its prices. These sources are "manual" or "prices.tf". These sources do not apply to custom auto-pricers.

**Describe the feature you'd like**
I propose a third price source, `custom`, to act as a differentiator for use in trade summaries, autokeys, and the like. The exact pricer or API is not required to be specified, simply `custom` is enough.

**Describe alternatives you've considered**
There are no alternatives to this enhancement that I could think of.

**Got any bounty as a reward for anyone fulfills your request?**
N/A

**Additional context**
Some examples in the code where "manual" or "prices.tf" need to be replaced to use "manual", "prices.tf" or "custom":
https://github.com/TF2Autobot/tf2autobot/blob/ab80ed170609baae457ba071ea89a1a094d266e5/src/classes/Commands/sub-classes/Manager.ts#L684
https://github.com/TF2Autobot/tf2autobot/blob/62ea10d2fd5dd5075932141478818a2633074870/src/classes/Commands/sub-classes/Help.ts#L62
https://github.com/TF2Autobot/tf2autobot/blob/ab80ed170609baae457ba071ea89a1a094d266e5/src/classes/MyHandler/offer/review/send-review.ts#L133
https://github.com/TF2Autobot/tf2autobot/blob/ab80ed170609baae457ba071ea89a1a094d266e5/src/lib/DiscordWebhook/sendOfferReview.ts#L123

This can be done in a way that specifies the current source globally. So the bot can easily refer to something such as `pricingSource`, and this would have either `manual`, `prices.tf` or `custom` so it can be easily referred to.

PR #364, opened 2021-02-23
fix typo, etc

idinium96:
- 🔄 replace bracket with just a dash to not confuse the generated URL (Steam Chat, on an error to accept an offer)
- 🔨 fix the wrong position (pure and inventory in Offer review - Discord Webhook)
- 🤝 custom-pricer friendly (#363)

### DIFF
--- a/src/classes/MyHandler/offer/accepted/updateListings.ts
+++ b/src/classes/MyHandler/offer/accepted/updateListings.ts
@@ -153,7 +153,9 @@ export default function updateListings(
                         `(+ ${priceFromOptions.keys > 0 ? `${pluralize('key', priceFromOptions.keys, true)}, ` : ''}${
                             priceFromOptions.metal
                         } ref)` +
-                        (isCustomPricer ? '' : `\nhttps://www.prices.tf/items/${sku}`);
+                        (isCustomPricer
+                            ? '\n - Base selling price was fetched from custom auto-pricer'
+                            : `\nhttps://www.prices.tf/items/${sku}`);
 
                     log.debug(msg);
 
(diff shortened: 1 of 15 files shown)


src/classes/MyHandler/offer/accepted/updateListings.ts
```diff
@@ -151,7 +153,7 @@ export default function updateListings(
                         `(+ ${priceFromOptions.keys > 0 ? `${pluralize('key', priceFromOptions.keys, true)}, ` : ''}${
                             priceFromOptions.metal
                         } ref)` +
-                        `\nhttps://www.prices.tf/items/${sku}`;
+                        (isCustomPricer ? '' : `\nhttps://www.prices.tf/items/${sku}`);
```
Bonfire, line 156: Maybe we can add something here instead of it being blank? Something to specify the price was fetched from the custom pricer.

Bonfire (approved):
Looks good to me. Thanks for the quick turnaround time!

src/classes/MyHandler/offer/accepted/updateListings.ts
idinium96: Oh sure thing, but I am not sure what's the best thing to put.

Bonfire: I'm thinking something like "Price fetched from custom auto-pricer"

SkiLEXx (approved):
All is well.

src/classes/MyHandler/offer/accepted/updateListings.ts
idinium96: Alright, will do that I guess.

PR #376, opened 2021-02-26
[fix] Build images only for Node.js LTS

rennokki:
I quickly realized that 15.10 should not be used since it's not LTS and `lts-alpine` is LTS. I decided to change the Docker images build for LTS only.

### DIFF
--- a/.github/workflows/docker-commit.yml
+++ b/.github/workflows/docker-commit.yml
@@ -19,12 +19,11 @@ jobs:
     strategy:
       matrix:
         node:
-          - 15.10-alpine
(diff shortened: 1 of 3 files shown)


idinium96:
Thank you. I thought you just gonna make it only `lts-alpine`, but I guess this is how things work. I will try to do some research on this, but anyway, let me merge this.

Btw, you're already a collaborator, meaning you do not need to use your forked repo. Clone this main repo on your local machine and create a new branch from `development`, do your changes, commit and push it, then create a pull request to `development` from your working branch. 🤗

PR #381, opened 2021-02-28
🔨 fix sending duplicated trade summary

idinium96:


arik123 (approved):
Everything seems to be in order.

Issue #326, opened 2021-02-10
Invalid item quantity in deposits if partner does not have item


Bonfire:
**Describe the bug**
Currently, if you attempt to deposit an invalid quantity of an item you do not have, the bot will attempt to process an offer with one less of the item and error.

**To Reproduce**

1. Deposit some quantity of an item you do not have
2. Send the `!checkout` command to execute the trade
3. Verify that the bot mentions you do not have the item, and offers to create a trade with one less of the item
4. Verify that the bot errors out when attempting to create the trade

**Expected behavior**
If possible, the bot should attempt to recognize that the user does not have the item. If not possible, the bot should at least not attempt to create a trade with one less of the quantity specified.

**Screenshots and logs**
![image](https://user-images.githubusercontent.com/5704760/107464333-12064300-6b2e-11eb-86b6-a3e2ea1277fa.png)
An example where my account does not have any refined metal, and I asked to deposit 5. The bot says it cannot find any, and attempts to send an offer for 4 refined metal.

**Additional context**
N/A

Version: TF2Autobot@v3.2.5

joekiller:
Thanks for this clear report

Issue #325, opened 2021-02-10
Perform partner item lookup on !deposit


Bonfire:
**Is your feature request related to a problem? Please describe.**
Currently, the bot does not check to see if you have an item (or a specific quantity of items) when depositing (using `!deposit`) until the cart starts being processed. This results in your having to re-adjust the cart which is a pain.

**Describe the feature you'd like**
This "does item/quantity exist" check is done when generating the trade offer, and I believe it should be done prior to that. However, I could see possible issues stemming from the fact that the bot may not have the most up-to-date inventory from the trade partner. This may need to be considered.

**Describe alternatives you've considered**
To verify that the exact item and quantity is provided when depositing items (which isn't very user-friendly!)

**Got any bounty as a reward for anyone fulfills your request?**
N/A

**Additional context**
![image](https://user-images.githubusercontent.com/5704760/107463750-c606ce80-6b2c-11eb-87ba-384b5ca37b3a.png)

Image showing me looking to deposit 999,999 refined metal (which clearly can't be done). Yet the bot happily goes along with it until the trade offer is actually made.

PR #382, opened 2021-02-28
🔨 improve `!deposit` command and perform items lookup

idinium96:
Solves [#325, #326]

image

### DIFF
--- a/src/classes/Commands/Commands.ts
+++ b/src/classes/Commands/Commands.ts
@@ -832,6 +832,19 @@ export default class Commands {
                 this.weaponsAsCurrency.enable ? this.bot.craftWeapons : [],
                 this.weaponsAsCurrency.enable && this.weaponsAsCurrency.withUncraft ? this.bot.uncraftWeapons : []
             );
+
+        if (amount > 0) {
+            const cartAmount = cart.getTheirCount(params.sku);
+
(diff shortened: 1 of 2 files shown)


idinium96:
Latest commit:

image

src/classes/Commands/Commands.ts
```diff
@@ -783,16 +832,25 @@ export default class Commands {
                 this.weaponsAsCurrency.enable ? this.bot.craftWeapons : [],
                 this.weaponsAsCurrency.enable && this.weaponsAsCurrency.withUncraft ? this.bot.uncraftWeapons : []
             );
+
+        if (amount > 0) {
+            const cartAmount = cart.getTheirCount(params.sku);
```
arik123, line 837: Why is there a possibility to have 0 amount. Shouldn't that be caught at the start and returned an error.

arik123 (changes requested):

src/classes/Commands/Commands.ts
idinium96: Because you can use a value less than 0 to remove things you've added to your cart.

https://github.com/TF2Autobot/tf2autobot/blob/a1497b54d2630b40ecbcd8cef7b8a8a4ab2d6400/src/classes/Commands/Commands.ts#L854

arik123 (approved):

PR #383, opened 2021-02-28
✨ option to disable socket entirely

idinium96:
New env variable: `ENABLE_SOCKET` (default is `true`)
- Set to `false` if you want to use the bot with everything in the pricelist to manually priced (disable socket means your bot will never update prices if the item `autoprice` parameter is set to `true` - Oh, your key rate will also stay the same unless you manually update it or restarted).


arik123 (approved):
Everything seems to be in order

PR #384, opened 2021-02-28
🔄 more flexibility when adding items with params

idinium96:
- at the same time, fix the bot crashed if the user inputs the quality parameter as a number.


arik123 (approved):
Everything seems to be fine.

arik123 (approved):

PR #385, opened 2021-02-28
✨ option to only show price update that is in stock

idinium96:


arik123 (approved):
Everything seems to be just fine

PR #387, opened 2021-03-01
🔨 refactor `!find` command

idinium96:
### Old:
-    Problem:
    - only show 100 entries if too many.
    - show in `/code` and entry form will cause the chat window to lag
-    Screenshot:
image

### New:
-    Only show `sku - name`
-    Can list all and no lagging issue.
image
.
.
.
image

### DIFF
--- a/src/classes/Commands/sub-classes/PricelistManager.ts
+++ b/src/classes/Commands/sub-classes/PricelistManager.ts
@@ -1289,22 +1289,37 @@ export default class PricelistManagerCommands {
         if (filterCount === 0) {
             this.bot.sendMessage(steamID, `No items found with ${display.join('&')}.`);
         } else {
-            this.bot.sendMessage(steamID, `Found ${pluralize('item', filterCount, true)} with ${display.join('&')}\n`);
             const list = filter.map(
                 (entry, i) => `${i + 1}. ${entry.sku} - ${this.bot.schema.getName(SKU.fromString(entry.sku))}`
             );
-
             const listCount = list.length;
-            const loops = Math.ceil(listCount / 200);
+
+            const limit =
+                params.limit === undefined ? 200 : (params.limit as number) <= 0 ? -1 : (params.limit as number);
+
+            this.bot.sendMessage(
+                steamID,
+                `Found ${pluralize('item', filterCount, true)} with ${display.join('&')}${
+                    limit !== -1 && params.limit === undefined && listCount > 200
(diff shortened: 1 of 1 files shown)


src/classes/Commands/sub-classes/PricelistManager.ts
```diff
@@ -1288,50 +1288,27 @@ export default class PricelistManagerCommands {
...
-                    `/code ${this.generateOutput(filter, true, 80, filterCount > 100 ? 100 : filterCount)}`
-                );
+
+            const listCount = list.length;
+            if (listCount > 200) {
+                const loops = Math.ceil(listCount / 200);
+
+                for (let i = 0; i < loops; i++) {
+                    const last = loops - i === 1;
+                    await sleepasync().Promise.sleep(1 * 1000);
```
arik123, line 1303: put this line after first iteration

src/classes/Commands/sub-classes/PricelistManager.ts
```diff
@@ -1288,50 +1288,27 @@ export default class PricelistManagerCommands {
...
-                await sleepasync().Promise.sleep(1 * 1000);
-                this.bot.sendMessage(steamID, `/code ${this.generateOutput(filter, true, 60, 80)}`);
-                await sleepasync().Promise.sleep(1 * 1000);
-                this.bot.sendMessage(
-                    steamID,
-                    `/code ${this.generateOutput(filter, true, 80, filterCount > 100 ? 100 : filterCount)}`
-                );
+
+            const listCount = list.length;
+            if (listCount > 200) {
```
arik123, line 1298: this path should be fine for any listing count

arik123 (changes requested):
There is no need for 2 different codepaths when sending messages,
put sleep after first message - faster response
I think that some kind of limit would be usefull, maybe add option `limit=<value>` and if it is set to -1 then it would list all. Also set some good default limit 50-200.

src/classes/Commands/sub-classes/PricelistManager.ts
idinium96: I see. So no need to do that if statement. got it. thanks

idinium96:
Lasted commit:
- Default limit = 200
- if result < 200:
image

- if result > 200 and limit not defined:
image

- if result > 200 and limit is defined:
image
image
image

arik123 (approved):

PR #395, opened 2021-03-03
⛔ don't execute if globalKeyPrices still undefined

idinium96:
- fixes:
image

- Occurred because the socket is running before we get the globalKeyPrices


Bonfire:
Appears to be a small change that adds one more failsafe to bot startup. All good with me. Thanks for fixing this so quickly!

Bonfire (approved):

PR #398, opened 2021-03-03
[3.x] Install pm2 in Docker

rennokki:


Dockerfile
```diff
@@ -6,9 +6,8 @@ LABEL maintainer="Renoki Co. <alex@renoki.org>"
 
 COPY . /app
 
-RUN npm install typescript@latest -g && \
+RUN npm install typescript@latest pm2 -g && \
```
idinium96, line 9: pm2 is installed, but will it run with it (the ENTRYPOINT)?
And also, maybe best use `pm2@latest`

arik123: it will not run with entrypoint, but that is probably not needed, it is there just to add the option to use it for people that want to

idinium96: alright noted.

PR #399, opened 2021-03-03
Refactor Bot.ts

arik123:
Cleaned up Bot.ts

There still is ESLINT error on L336, which could be fixed by making `onPricelist` synchronous.

### DIFF
--- a/src/classes/Bot.ts
+++ b/src/classes/Bot.ts
@@ -106,7 +106,7 @@ export default class Bot {
 
     private ready = false;
 
-    constructor(public botManager: BotManager, public options: Options, private priceSource: Pricer) {
+    constructor(public readonly botManager: BotManager, public options: Options, private priceSource: Pricer) {
         this.botManager = botManager;
 
         this.schema = this.botManager.getSchema;
--- a/src/classes/Bot.ts
+++ b/src/classes/Bot.ts
@@ -106,7 +106,7 @@ export default class Bot {
 
     private ready = false;
 
-    constructor(botManager: BotManager, public options: Options, private priceSource: Pricer) {
+    constructor(public botManager: BotManager, public options: Options, private priceSource: Pricer) {
         this.botManager = botManager;
 
         this.schema = this.botManager.getSchema;
--- a/src/classes/Bot.ts
+++ b/src/classes/Bot.ts
@@ -333,7 +333,7 @@ export default class Bot {
 
         this.addListener(this.listingManager, 'heartbeat', this.handler.onHeartbeat.bind(this), true);
 
-        this.addListener(this.pricelist, 'pricelist', this.handler.onPricelist.bind(this.handler), false);
+        this.addListener(this.pricelist, 'pricelist', void this.handler.onPricelist.bind(this.handler), false);
         this.addListener(this.pricelist, 'price', this.handler.onPriceChange.bind(this.handler), true);
 
         return new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


idinium96:
I will let sir @joekiller review this since he made these changes.

joekiller:
Hi @arik123 if you bind the way it is setup then each binding is an anonymous function created at that moment of the method call and this can have unintended side effects. For instance if the bot restarts after a crash then you will get double the events. If it crashed again, it would setup triple events. Also because the functions are anonymous if you call removeAllListeners on the socket, it will not be able to remove those functions.

If I am mistaken please let me know. Otherwise I would reject this.

arik123:
> Hi @arik123 if you bind the way it is setup then each binding is an anonymous function created at that moment of the method call and this can have unintended side effects. For instance if the bot restarts after a crash then you will get double the events. If it crashed again, it would setup triple events. Also because the functions are anonymous if you call removeAllListeners on the socket, it will not be able to remove those functions.
> 
> If I am mistaken please let me know. Otherwise I would reject this.

Is there a way how I reproduce this behaviour ? And is there a way how to debug with PM2 ?

joekiller:
I think you could just call !restart several times and then you'll see
duplicate price update events and such.

Sent via mobile

idinium96:
You can put this in your `.env` to see socket activities (credit to @joekiller):
image

joekiller:
Read this: https://medium.com/@DavideRama/removeeventlistener-and-anonymous-functions-ab9dbabd3e7b

arik123:
that should not be a problem, because there does not seem to be any removeListener used

arik123:
I tried restarting the bot, but it did not seem to do anything as it should be.

joekiller:
To shutdown the bot cleanly the socket manager shutdown call uses this.
socket.removeAllListeners()
https://github.com/TF2Autobot/tf2autobot/blob/master/src/classes/BotManager.ts#L244

So it is used.
https://github.com/TF2Autobot/tf2autobot/blob/master/src/classes/MyHandler/SocketManager.ts#L70

One could argue that only a subset of the functions actually need to be
bound to a named instance and I would support that refactor if you want but
anything that gets bound to socket events need to have named instances not
anonymous functions.

Sent via mobile

arik123:
Even if `removeAllListeners()` was somehow unable to remove anonymous listener, if you restart the bot, there should be no problem with double registration of handlers.
Also as far as I can tell, the only two cases, when events are bound to socketManager are [Pricelist](https://github.com/TF2Autobot/tf2autobot/blob/95c61995bd80de23e1bf2754a57e44be23fd7d7b/src/classes/Pricelist.ts#L197) and inside [SocketManager](https://github.com/TF2Autobot/tf2autobot/blob/95c61995bd80de23e1bf2754a57e44be23fd7d7b/src/classes/MyHandler/SocketManager.ts#L47-L61) itself, and in both cases there are no anonymous functions

idinium96:
I'd prefer best practices.

arik123:
and what are the best practises ?

idinium96:
> and what are the best practises?

from the reference provided by @joekiller and his further explanations?

joekiller:
Okay I think we encapsuled the part that could blow up in the socket
manager so I'll pass this

Sent via mobile

idinium96:
alright, now let me see the cause of the error.

joekiller:
But make sure you fix the linting problem which may require a named
reference or some syntax update to bind to the async functions

Sent via mobile

idinium96 (approved):
Since @joekiller approves, I approve.

idinium96 (approved):
merging...

PR #412, opened 2021-03-04
Add option toggle autoAddInvalidUnusual = false

joekiller:
If the bot accepts an Unusual item that is not on the price list it will not put it back up for sale unless autoAddInvalidUnusual is set to true. The option is set to false by default to protect users who use the generic price SKU and unaware that autoAddInvalidItems would sell off anything purchased via that generic sku.

Based off conversation in #400

### DIFF
--- a/src/classes/MyHandler/offer/accepted/updateListings.ts
+++ b/src/classes/MyHandler/offer/accepted/updateListings.ts
@@ -30,6 +30,7 @@ export default function updateListings(
     const inventory = bot.inventoryManager.getInventory;
     const hv = highValue.items;
     const normalizePainted = opt.normalize.painted;
+    const dwEnabled = opt.discordWebhook.sendAlert.enable && opt.discordWebhook.sendAlert.url !== '';
 
     for (const sku in diff) {
         if (!Object.prototype.hasOwnProperty.call(diff, sku)) {
@@ -71,11 +72,19 @@ export default function updateListings(
             !isAdmin &&
             opt.pricelist.autoAddInvalidItems.enable;
 
-        const receivedNotInPricelist =
+        const receivedHighValueNotInPricelist =
             !existInPricelist &&
             isNotPureOrWeapons &&
             isNotSkinsOrWarPaint && // exclude War Paint (could be skins)
             isDisabledHV && // This is the only difference
+            !isAdmin;
+
+        const receivedUnusualNotInPricelist =
+            !existInPricelist &&
+            isNotPureOrWeapons &&
+            isNotSkinsOrWarPaint &&
+            item.quality === 5 &&
+            opt.pricelist.autoAddInvalidUnusual === false &&
             !isAdmin;
 
         const isAutoDisableHighValueItems =
@@ -173,7 +182,7 @@ export default function updateListings(
                     log.debug(msg);
 
                     if (opt.sendAlert.enable && opt.sendAlert.autoAddPaintedItems) {
-                        if (opt.discordWebhook.sendAlert.enable && opt.discordWebhook.sendAlert.url !== '') {
+                        if (dwEnabled) {
                             sendAlert('autoAddPaintedItems', bot, msg.replace(/"/g, '`'));
                         } else {
                             bot.messageAdmins(msg, []);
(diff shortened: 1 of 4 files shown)


src/classes/MyHandler/offer/accepted/updateListings.ts
```diff
@@ -44,6 +44,9 @@ export default function updateListings(
         const isDisabledHV = highValue.isDisableSKU.includes(sku);
         const isAdmin = bot.isAdmin(offer.partner);
         const isNotSkinsOrWarPaint = SKU.fromString(sku).wear === null;
+        // if item is unusual and autoAddInvalidUnusual is set to true then we allow addInvalidUnusual.
+        // If the item is not an unusual sku, we "allow" still (no-op)
+        const addInvalidUnusual = SKU.fromString(sku).quality === 5 ? opt.pricelist.autoAddInvalidUnusual : true;
```
idinium96, line 49: This is mainly for the Generic Unusual feature right?
So I guess it's much better to do it like this (together with some small refactorization):

```ts
const item = SKU.fromString(sku);
const isNotSkinsOrWarPaint = item.wear === null;
const addInvalidUnusual = item.quality === 5 && item.effect === null ? opt.pricelist.autoAddInvalidUnusual : true;
```

src/classes/Options.ts
```diff
@@ -83,6 +83,9 @@ export const DEFAULTS = {
         autoAddInvalidItems: {
             enable: true
         },
+        addInvalidUnusual: {
```
idinium96, line 86: rename this to `autoAddInvalidUnusual` will fix the test error.

idinium96 (changes requested):

src/classes/MyHandler/offer/accepted/updateListings.ts
idinium96: oh, wait. I don't think `item.effect === null` is correct.

idinium96: Yeah, scratch that. My bad.

joekiller: yeah I was about to say. I think in this case the SKU detected would be the SKU of the actual item at this point, so it wuld have an effect

idinium96: Yep correct. It's not like when receiving the painted item (with `normalize.painted.their` set to `false`)

src/classes/MyHandler/offer/accepted/updateListings.ts
```diff
@@ -36,14 +36,18 @@ export default function updateListings(
...
         const inPrice = bot.pricelist.getPrice(sku, false);
 
         const existInPricelist = inPrice !== null;
         const isDisabledHV = highValue.isDisableSKU.includes(sku);
         const isAdmin = bot.isAdmin(offer.partner);
-        const isNotSkinsOrWarPaint = SKU.fromString(sku).wear === null;
+        const isNotSkinsOrWarPaint = item.wear === null;
+        // if item is unusual and autoAddInvalidUnusual is set to true then we allow addInvalidUnusual.
+        // If the item is not an unusual sku, we "allow" still (no-op)
+        const addInvalidUnusual = item.quality === 5 ? opt.pricelist.autoAddInvalidUnusual : true;
```
joekiller, line 50: this is somewhat a strange statement but it really should be "true" if either:
1. the item is unsual and we autoAddInvalidUnusual or,
2. if the item isn't an unusual.

src/classes/MyHandler/offer/accepted/updateListings.ts
joekiller: this is somewhat a strange statement but it really should be "true" if either:

the item is unsual and we autoAddInvalidUnusual or,
if the item isn't an unusual.

idinium96 (commented):
> this is somewhat a strange statement but it really should be "true" if either:

> - the item is unsual and we autoAddInvalidUnusual or,
> - if the item isn't an unusual.

very true. but will the bot notify the owner about this?

joekiller:
Not at the moment. That could be considered a secondary part to this safety net. An option like notifyAcceptedInvalidUnusual or something?

idinium96:
yeah, something like that. I might add that while you're focusing on your medical work.

idinium96:
The latest commit should do the notifying job.

idinium96:
this looks good now I guess. merging...

PR #413, opened 2021-03-04
[3.x] HTTP API healthcheck

rennokki:
The concept of having an API for the bot can ease multiple tasks:

- ~sending commands to the bot via HTTP (opens up new ways of implementing SDKs for tf2autobot, implementable in any language since it's REST)~
- ~reading information about the BOT, including current item configurations for trading~
- periodic healthchecks performed on the bot via HTTP endpoint (useful for Docker/Kubernetes, but not limited to)

### Roadmap
- [x] Base implementation, `/health` endpoint for checking readiness/aliveness
- [x] Uptime endpoint for reading the current uptime status
- [ ] ~Configured items/bots endpoint for reading the configured items and bots~
- [ ] ~Tests, lots of tests~

[The GUI v3 rework already has an entrypoint for the API](https://github.com/TF2Autobot/tf2autobot-gui/blob/rework-V3), so sticking just with healthchecks and uptime should be fine. 👍🏼


idinium96:
Love the idea. I will let the others share their thoughts too.

arik123:
I'm working on something simular - but I'm using mode-IPC for it, is gonna be used for GUI

rennokki:
Discussed with @arik123 earlier on Discord about IPC and the need of having an API server on a bot. The main points were:

1. Having a bot running is not going to start the second one since the API is already running on the specified port. This isn't an issue on Docker or any other distributed system since the API is isolated from external networking, and will only be accessible at their respective container names.
2. The GUI is on its way and will act as a master for all bot processes. This is an excellent idea because it doesn't expose the bot on the network and the other functionalities on this PR are redundant, so I put a strike on them on the roadmap.

The only left issue is the healthcheck functionality. The healthcheck can pass/fail if I either:
- call an URL present on the process (200 = pass, non-200 = fail)
- ping a TCP socket (established = pass, rejected = fail)
- execute a command (exit code zero = pass, exit code non-zero = fail)

I was looking for the TCP or the command execs for healthchecks, but I'm not sure which one to pick or how we can acheive this. The HTTP `/health` check is still good, but it doesn't work for simple processes, outside Docker/Kubernetes, excepting if a different port is set for each one.

idinium96:
Thanks for the update. You may keep this pull request open so every progress/discussion can be seen here.
This may sound weird but a built-in GUI (not enabled by default) might be good, but not sure if that will use a lot of resources.

arik123:
it could be probably implemented such that, it would take near to nothing if not enabled

idinium96:
> it could be probably implemented such that, it would take near to nothing if not enabled, but then the whole part of GUI that I have planned - Managment of multiple bots would not work

Oh right true. in that case, probably best to make the GUI a separate app.

rennokki:
Maybe the API can still exist (and be disabled by default) for non-GUI-based applications, like managing your own bots?

idinium96:
> Maybe the API can still exist (and be disabled by default) for non-GUI-based applications, like managing your own bots?

Yes, sure thing.

src/app.ts
```diff
@@ -130,4 +133,10 @@ void botManager.start(options).asCallback(err => {
     if (err) {
         throw err;
     }
+
+    void httpManager.start().asCallback(err => {
+        if (err) {
+            throw err;
+        }
+    });
 });
```
idinium96, line 142: I think might be better to never initiate HttpManager class from the beginning if it's disabled.

maybe like:
```ts
void botManager.start(options).asCallback(err => {
    if (err) {
        throw err;
    }
    
    if (options.enableHttpApi) {
        const httpManager = new HttpManager(options);

        void httpManager.start().asCallback(err => {
            if (err) {
                throw err;
            }
        }
    });
});
```

src/app.ts
```diff
@@ -4,6 +4,7 @@ const { version: BOT_VERSION } = require('../package.json');
 import { getPricer } from '@pricer/pricer';
 import Pricer, { GetPricerFn } from './classes/Pricer';
 import { loadOptions } from './classes/Options';
+import HttpManager from './classes/HttpManager';
```
arik123, line 7: What about importing only if enabled, to reduce memory usage slightly, but this is not that important

src/app.ts
rennokki: https://github.com/TF2Autobot/tf2autobot/pull/413/commits/6a67423fe3df5358d4ef0000fce7723e60699248

src/classes/HttpManager.ts
```diff
@@ -0,0 +1,54 @@
...
+export default class HttpManager {
+    /**
+     * The Express.js app.
+     */
+    protected app: express.Application;
+
+    /**
+     * The Express.js server app.
+     */
+    public server;
```
arik123, line 20: Don't see any uses for this variable

arik123 (commented):
What about setting default port to something random, so there is smaller chance of collision with other programs.

arik123:
Also maybe add some note to the source, that tells everyone about GUI as main API provider.

src/app.ts
rennokki: ES6 and TS don't allow importing with `import from` within classes.

arik123: I know, either use `require` or let it be like it is now

arik123 (approved):

idinium96:
Now let say I want to use this and want to check by bot uptime via an API, and I am running my bot on a VPS, I need to type this on my browser URL:
`http;//<IP_ADDRESS:PORT>/uptime`

right?

rennokki:
@idinium96 Yes.

idinium96 (approved):
roger that 👍

merging before going to bed...

PR #420, opened 2021-03-06
🔨 refactor: use Map and Set

idinium96:
- much better right?


Bonfire (approved):
Appears to be the correct usage. Approved.

PR #458, opened 2021-03-12
Fix generic without generic

### LINKED ISSUE #457: Unusuals: 🟦_OVERSTOCKED while stock < max
**Describe the bug**
Bot holds correct trade offers for review with reason `🟦_OVERSTOCKED`, while current stock < max.
This only happens with Unusuals, and only when there are several of the same item with different effects in the backpack.
For example, when:
- I am buying Massed Flies Dread Knot for 6 keys
- My current stock for this item is 0
- My max stock for this item is 2
- I currently own > 2 Unusual Dread Knots with different effects

This bug only happens on received trade offers.
Everything works as intended when seller trades via chat message.

**To Reproduce**

1. Create buy order for Unusual item with effect [X] and set maximum stock to [N]. _(User can choose X & N)_
2. Have less than [N] of said item with effect [X] available in backpack.
3. Have more than [N] of said item with different effects available in backpack.
4. Wait for seller to send a correct trade offer to the bot containing Unusual item with effect [X].
5. Bot will mark offer as `🟦_OVERSTOCKED` and hold the trade for manual review.

**Expected behavior**
Bot accepts correct trade offers containing Unusual items when stock < max.

**Screenshots and logs**
Trade below shows current stock = 0 and max stock = 2 for the Infernal Smoke Bad Pipes, 
yet the offer is marked as `🟦_OVERSTOCKED`.

Bad Pipes with different effects available in the backpack.

Version: 
3.6.5

joekiller:
Only use generic counts if we have a generic sku in the first place.

fixes #457

### DIFF
--- a/src/classes/InventoryManager.ts
+++ b/src/classes/InventoryManager.ts
@@ -1,6 +1,7 @@
 import Currencies from 'tf2-currencies-2';
 import Inventory from './Inventory';
 import Pricelist from './Pricelist';
+import SKU from 'tf2-sku-2';
 
 export default class InventoryManager {
     constructor(private readonly pricelist: Pricelist, private inventory: Inventory = null) {
@@ -39,18 +40,32 @@ export default class InventoryManager {
         }
 
         let genericCheck = generics;
+        let genericIndex = -1; // index of a generic match
         // if we looking at amount we can trade and the sku is a generic unusual, always set generic to true
         const isGenericSku = /^[0-9]*;5$/.test(sku);
         if (isGenericSku) {
             genericCheck = true;
         }
 
+        const gSku = SKU.fromString(sku);
+        gSku.effect = null;
+        if (generics) {
+            // figure out if we even have a generic sku
+            // Index of of item in Pricelist
+            genericIndex = this.pricelist.getIndex(null, gSku);
+        }
+
         // Pricelist entry
-        const match = genericCheck ? this.pricelist.getPrice(sku, true, true) : this.pricelist.getPrice(sku, true);
+        const match =
+            genericCheck && genericIndex !== -1
+                ? this.pricelist.getPrice(SKU.fromObject(gSku), true, true)
+                : this.pricelist.getPrice(sku, true);
 
         // Amount in inventory should only use generic amount if there is a generic sku
         const amount =
-            genericCheck && match ? this.inventory.getAmountOfGenerics(sku, true) : this.inventory.getAmount(sku, true);
+            genericCheck && match && genericIndex !== -1
+                ? this.inventory.getAmountOfGenerics(SKU.fromObject(gSku), true)
+                : this.inventory.getAmount(sku, true);
 
         if (match === null) {
             // No price for item
--- a/src/classes/InventoryManager.ts
+++ b/src/classes/InventoryManager.ts
@@ -54,10 +54,11 @@ export default class InventoryManager {
             // Index of of item in Pricelist
             genericIndex = this.pricelist.getIndex(null, gSku);
         }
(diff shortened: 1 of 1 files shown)


idinium96:
with these logic changes, which price will the bot use? A buy order with generic or non-generic?

src/classes/InventoryManager.ts
```diff
@@ -39,18 +40,32 @@ export default class InventoryManager {
...
         if (isGenericSku) {
             genericCheck = true;
         }
 
+        const gSku = SKU.fromString(sku);
+        gSku.effect = null;
+        if (generics) {
+            // figure out if we even have a generic sku
+            // Index of of item in Pricelist
+            genericIndex = this.pricelist.getIndex(null, gSku);
```
idinium96, line 55: cont.
Because if got both generic and non-generic in pricelist, this will always > -1

joekiller: amount can buy, generic max still overrides normal.I think you are right. I have made it such that it checks for both a generic and a normal sku. If it finds the normal sku, even in generic mode, it will use the normal sku for price. For

src/classes/InventoryManager.ts
```diff
@@ -39,18 +40,33 @@ export default class InventoryManager {
...
+            genericCheck && genericIndex !== -1 && normalIndex === -1
+                ? this.pricelist.getPrice(SKU.fromObject(gSku), true, true)
+                : this.pricelist.getPrice(sku, true);
 
         // Amount in inventory should only use generic amount if there is a generic sku
         const amount =
-            genericCheck && match ? this.inventory.getAmountOfGenerics(sku, true) : this.inventory.getAmount(sku, true);
+            genericCheck && match && genericIndex !== -1
+                ? this.inventory.getAmountOfGenerics(SKU.fromObject(gSku), true)
+                : this.inventory.getAmount(sku, true);
```
idinium96, line 69: looks good to me, but I wonder why that `normalIndex` is not here as well.

idinium96: because let's say this is `false` (due to `normalIndex !== -1`)
https://github.com/TF2Autobot/tf2autobot/blob/fc585b3a395affcaef5d77c4d15634525565f3f0/src/classes/InventoryManager.ts#L61
It should then read this
https://github.com/TF2Autobot/tf2autobot/blob/fc585b3a395affcaef5d77c4d15634525565f3f0/src/classes/InventoryManager.ts#L63

but the `this.pricelist.getPrice(sku, true)` return non-null result (which mean non-generic exist in the pricelist).

Then, when it read this
https://github.com/TF2Autobot/tf2autobot/blob/fc585b3a395affcaef5d77c4d15634525565f3f0/src/classes/InventoryManager.ts#L67
it will be `true` and use `getAmountOfGenerics` instead of `getAmount` for non-generic.

Correct me if I am wrong.

idinium96: oh wait, it can be `false` due to `genericIndex === -1`. right?

joekiller: The first statement says, only pull a generic sku if we first, have a generic sku and second, we don't have a specific sku for the item. IE, generic sku never overrides price settings for specific.

The second statement says, only use the "generic can buy amount" if there is a generic sku present. The second really fixes the present bug because prior it was always trying to use the generic amount even if a generic sku didn't exist so it would get hits on other particle types even though there wasn't a generic sku.

idinium96: got it. Thanks for the reply. I will merge this soon.

idinium96 (approved):

PR #461, opened 2021-03-14
🔨⚡moderate refactor, slightly improve performance

idinium96:
- `~src/classes/`:
    - `Carts/UserCart.ts`:
        - get rid of `highValueInput`.
    - `Listings.ts`:
        - get rid of "Checking account info" logs.
        - get rid of `alreadyMakingAListing` check - unnecessary and might cause different price problems.
    - `MyHandler/MyHandler.ts`:
        - reposition `filterReasons` to be outside MyHandler class.
        - get rid of `highValueInput`.
        - use a better method to check full uses for Dueling Mini-Game and/or Noise Maker.
        - get rid of `notIncludeCraftweapons` - unnecessary.
        - get rid of `recheckMatch` and `isCanBePriced` - unnecessary since the `match` already with `onlyEnabled = false`.
        - remove the old overstocked debugging log.
        - only execute Invalid Value Exception check if `exceptionSKU` array is not 0.
        - remove the condition in `sortInventory`.
    - `MyHandler/offer/accepted/processAccepted.ts`:
        - replace `offerSent` with `highValue`
    - `Pricelist.ts`:
        - replace `name` with `sku` on `updateOldPrices` method.
    - `Trades.ts`:
        - save `highValue` if defined.
- `~src/lib/`:
    - `DiscordWebhook/pricelistUpdate.ts`:
        - convert the function into const.
    - `DiscordWebhook/sendTradeSummary.ts`:
        - only execute check mention on sku if opt array length is not 0.
    - `extend/item/getTag.ts` → `extend/item/getItemTag.ts` (solve FIXME in `@tf2autobot/tradeoffer-manager` index.d.ts file )


arik123 (approved):
I don't see any problems.

PR #465, opened 2021-03-15
🚧 add a failsafe in `handlePriceChange` method

idinium96:
- prevent bot crashed:
image


Bonfire (approved):
Looks good to me. I'd prefer the admin to be notified if there was a failure to price an item that way they can investigate, but that is most likely too spammy. I think the logging here is a good middle-ground.

idinium96:
> Looks good to me. I'd prefer the admin to be notified if there was a failure to price an item that way they can investigate, but that is most likely too spammy. I think the logging here is a good middle-ground.

Yea that's true. Not sure if sending unmentioned notification on the same webhook URL as the price update would be better or not.

idinium96:
I am adding it

PR #480, opened 2021-03-18
✨ add an option to not show owner's nickname

Dex-Spirit:
Option to hide owner's nickname on `!message` & `!accept`/`!decline`(if reply is included) commands.
Default is true (show nickname).


idinium96 (approved):
Thank you 👍

PR #497, opened 2021-03-25
PPU: 🔨 fix the unintended rare occurrence

idinium96:
Fix bot does not apply PPU when this occurs:
- The bot bought for 16 keys (which is about ~7552 in scrap value at 52.44 ref/key):
image

- And then the price got updated with 14 keys, 21 ref / 15 keys, 52.44 ref - Which is for the selling price is basically the same as the previous buying price (broken prices.tf 😭).
image

And thus the item was sold for 0 profit.
image


SkiLEXx (approved):
Now this is a commit even I can comprehend :P

PR #503, opened 2021-03-26
✨ implement queue on `!pricecheckall` command

idinium96:
Prevent "Temporary IP banned" issue.
image

### DIFF
--- a/src/classes/Commands/Commands.ts
+++ b/src/classes/Commands/Commands.ts
@@ -249,7 +249,7 @@ export default class Commands {
         } else if (command === 'pricecheck' && isAdmin) {
             this.request.pricecheckCommand(steamID, message);
         } else if (command === 'pricecheckall' && isAdmin) {
-            void this.request.pricecheckAllCommand(steamID);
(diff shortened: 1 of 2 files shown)


idinium96:
The current implementation will cause the bot to use RAM quite a lot if there's any incoming offer or the bot is creating an offer while the pricecheck all is running in the background (refer to process name **Mai** in the screenshots below):
image
image

Is it because of this?
https://github.com/TF2Autobot/tf2autobot/blob/22d3ac5ed2026f40ccd68193a1fbd5ba4e8e668f/src/classes/Commands/sub-classes/Request.ts#L196
https://github.com/TF2Autobot/tf2autobot/blob/22d3ac5ed2026f40ccd68193a1fbd5ba4e8e668f/src/classes/Commands/sub-classes/Request.ts#L320-L322

idinium96:
Maybe instead of putting the entire class, we just put `true` in it?

idinium96:
Well nvm. I guess that huge RAM usage was caused by my current polldata.json file (which is currently already 15 MB)

PR #520, opened 2021-03-28
🩹 improvement on PPU (#506)

### LINKED ISSUE #506: Improve partialPriceUpdate while keeping work to a minimum
The root of the problem is we don't keep track for how much we buy and sell individual items. I am fully aware solving this root would require **a lot** of time, time which I am incapable of compensating anyone for. This is why I am proposing another addition to our current band-aid - `partialPriceUpdate` (abbreviation: PPU)

Current implementation of  PPU is not perfect. Its goal is to make sure we don't sell at a loss. 
In its current implementation it's still missing an important feature: **making sure we sell an item for more than we bought it for, no matter if it's 0.11 ref or 30 ref**

To make this feature work with as little extra work as possible we need to base everything around the idea that a single type of item (e.g. Non-Craftable Tour of Duty Ticket) has a maximum stock of 1. This means we only care for one side of the autoprice (either buying price **OR** selling price).

With this in mind I would like to make an example comparing current implementation vs. my proposed addition.

**Situation:**
Autoprice is currently **buying for 20 ref** and **selling for 25 ref**

`We just bought a Non-Craftable Tour of Duty Ticket (ToD) for 20 ref. 
Our goal is to sell this ToD for a profit, no matter if it's 5 ref or 1 ref.`

Let's reiterate how the current implementation of PPU handles this situation:
1. It will keep updating the buying price, even though it wouldn't matter since we aren't buying any more units of this item.

2. It will only update the selling price if it's above the current selling price, even though we could still be earning a positive return as long as the **updated selling price > the buying price for which we bought the item**

**Observations:**
1. We can **keep the buying price static**(don't update it) for as long as we didn't sell the item. Why? Because our maximum stock is 1 and we are not buying any more of this item.
=> We just tracked the buying price of an item. This means we can now use it to make sure we don't sell an item for less than we bought it for.

2. We aren't updating our selling price even when we could still earn a positive return.

**My proposition and example**
Let's take another look at my ToD example and see how great my addition would be.

`We just bought a Non-Craftable Tour of Duty Ticket (ToD) for 20 ref. 
Our goal is to sell this ToD for a profit, no matter if it's 5 ref or 1 ref. `

The first thing my addition would do is stop updating the buying price (here: 20 ref).
Now the bot knows the only acceptable selling price updates are above 20 ref if we wish to be earning a positive return.

Whenever it receives an updated selling price it checks whether it's above the static buying price. 
If this is the case it will update our selling price. 
If not it will keep the current selling price.

Once the item sells it updates both selling and buying price to make sure we are correctly pricing our items. 

Pricing manually but that is not why we use the bot.

**Got any bounty as a reward for anyone fulfills your request?**
None (for now)

idinium96:
Closes #506 

### New pricelist entry key (or parameter):
- `isPartialPriced` - https://github.com/TF2Autobot/tf2autobot/pull/525
- will ensure the `group` parameter stays intact.

### New options added:
- `sendAlert.partialPrice.onResetAfterThreshold` (default is `true`).
- `sendAlert.partialPrice.onBulkUpdatePartialPriced` (default is `true`).

### General Information 
- The new logic will **not** update the current buying price if stock exists.
- Since the bot does not get the last bought price, setting the current buying price to static will ensure that we never lose any profit.
- The bot will update only the selling price if the new selling price is more than the current static buying price OR the new selling price is more than the current selling price, but if these two conditions are not met, the new selling price will be the current static buying price + 0.11 ref (minimum profit).

### Codes
A. On price change/update old prices
https://github.com/TF2Autobot/tf2autobot/blob/4ea8c1c56d728236a07fc0dc0eebb2e765d1ca14/src/classes/Pricelist.ts#L1000-L1097

B. On item sold out
https://github.com/TF2Autobot/tf2autobot/blob/4ea8c1c56d728236a07fc0dc0eebb2e765d1ca14/src/classes/MyHandler/offer/accepted/updateListings.ts#L125-L130
https://github.com/TF2Autobot/tf2autobot/blob/4ea8c1c56d728236a07fc0dc0eebb2e765d1ca14/src/classes/MyHandler/offer/accepted/updateListings.ts#L347-L402

### Codes crackdown:
- A. On price change/update old prices
    - `(1)` The bot will first determine if this option is enabled, and that particular item is (i) currently in stock, (ii) does not exceed the time threshold, (iii) is not excluded item(s), **AND** (iv) the max is only set to 1:
https://github.com/TF2Autobot/tf2autobot/blob/4ea8c1c56d728236a07fc0dc0eebb2e765d1ca14/src/classes/Pricelist.ts#L1009-L1013
https://github.com/TF2Autobot/tf2autobot/blob/4ea8c1c56d728236a07fc0dc0eebb2e765d1ca14/src/classes/Pricelist.ts#L1027

        - `(1.1)` if `(1)` is `true`, the bot will then determine if (i) the item already has `isPartialPriced` set to `true`, (ii) the difference between the new selling price and the current buying price is negative or zero, **OR** (iii) the buying price changed:
https://github.com/TF2Autobot/tf2autobot/blob/4ea8c1c56d728236a07fc0dc0eebb2e765d1ca14/src/classes/Pricelist.ts#L1037-L1038
https://github.com/TF2Autobot/tf2autobot/blob/4ea8c1c56d728236a07fc0dc0eebb2e765d1ca14/src/classes/Pricelist.ts#L1050

            - `(1.1.1)` if `(1.1)` is `true`, then the bot will check if (i) the new selling price is more than the current buying price, **OR** (ii) the new selling price is more than the current selling price. If `true`, then we **update ONLY the selling price.**
https://github.com/TF2Autobot/tf2autobot/blob/4ea8c1c56d728236a07fc0dc0eebb2e765d1ca14/src/classes/Pricelist.ts#L1051-L1053
            - `(1.1.2)` else if `(1.1.1)` is `false`, the bot will update the selling price with an addition of 1 scrap from the buying price (the minimum profit).
https://github.com/TF2Autobot/tf2autobot/blob/4ea8c1c56d728236a07fc0dc0eebb2e765d1ca14/src/classes/Pricelist.ts#L1054-L1057
        - `(1.2)` else if `(1.1)` is `false`, if the item `isPartialPrice` is `false`, the bot will update the prices as usual.
https://github.com/TF2Autobot/tf2autobot/blob/4ea8c1c56d728236a07fc0dc0eebb2e765d1ca14/src/classes/Pricelist.ts#L1071-L1078
    - `(2)` else if `(1)` is `false`, the bot will first check if that particular item's `isPartialPriced` is `false`, OR if it's `true`, then check if it's already exceeded the time threshold OR no longer in stock. If everything `true`, the bot will update the prices as usual and reset `isPartialPriced` to `false`.
https://github.com/TF2Autobot/tf2autobot/blob/4ea8c1c56d728236a07fc0dc0eebb2e765d1ca14/src/classes/Pricelist.ts#L1080-L1093

B. On item sold out
    - When that particular item is sold out, the bot will update the buying and selling prices to the latest prices.tf or custom pricer, and `isPartialPriced` parameter will be reset to `false`.

#### 1
- The bot bought for 18.66 ref:
image

- The PPU in action:
image

- Loss prevented by selling it at 18.77 ref instead of the current selling price of 16.22 ref:
image

#### 2 (updated)
image

image

image


idinium96:
image

PR #568, opened 2021-04-11
[WIP] - v4

idinium96:


idinium96:
For some unknown reason, the `await prices.getSchema()` here:
https://github.com/TF2Autobot/tf2autobot/blob/d5ee6baee5f69bef28b036f4dd7d7f823740a8fb/src/classes/__tests__/Pricelist.ts#L17

only return the schema for Mann Co. Supply Crate Key:
image

And thus the test failed because the `raw` should have everything in the schema, not just key.

Issue #582, opened 2021-04-20
Option to see summaries of declined trades


RobotoLev:
**Is your feature request related to a problem? Please describe.**

I've started my own bot some days ago. I'm always frustrated when it declines trades, but I don't know why those trades were declined.

**Describe the feature you'd like**

I think, it may be good (especially for newbies) to have option for bot to send to Discord or Steam chat information about declined trades. It can be, for example, Discord webhook to channel where will be only summaries for not-accepted trades. And, next to it, option to disable it (then bot won't send this summaries, right how it's going now).

**Describe alternatives you've considered**

For now I have only one solution to understand why bot declined this or another trade. In this case admin needs to go to server and read logs to find information about that trade. It's may be hard in some times (for ex. when you use your phone to chat with bot).

**Got any bounty as a reward for anyone fulfills your request?**

N/A

**Additional context**

People may have problems with perfect configuration of bot and pricelist at first times. This feature will simplify debug process.

idinium96:
Thanks for creating an issue about this. I've already created a poll on [TF2Autobot Discord server](https://discord.gg/D2GNnp7tv8) about this 3 months ago but still not implemented because it's quite a headache to do it. Maybe will be implemented together with v4 later.

PR #624, opened 2021-05-09
Add Trade Declined Summary

Preport:
adds ability to receive declined trade summaries.

Sends Discord Webhook message or Steam message on declined offers.
Uses webhook `trade-declined`.
Message includes the reason and the description about why the trade was declined.
Adds new option to `options.json` called tradeDeclined

Fixes: https://github.com/TF2Autobot/tf2autobot/issues/582

### DIFF
--- a/src/classes/Options.ts
+++ b/src/classes/Options.ts
@@ -115,6 +115,7 @@ export const DEFAULTS: JsonOptions = {
     },
 
     tradeSummary: {
+        declinedTrade:{enable:false},
         showStockChanges: false,
         showTimeTakenInMS: false,
         showDetailedTimeTaken: true,
@@ -392,7 +393,6 @@ export const DEFAULTS: JsonOptions = {
         avatarURL: '',
         embedColor: '9171753',
         tradeSummary: {
-            declinedTrade:{enable:false},
             enable: true,
             url: [],
             misc: {
@@ -1123,7 +1123,6 @@ export interface TradeSummary {
     customText?: TradeSummaryCustomText;
 }
 
-
 interface TradeSummaryCustomText {
     summary: SteamDiscord;
     asked: SteamDiscord;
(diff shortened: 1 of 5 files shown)


idinium96:
Awesome! Thank you.
I will go through this PR sometime today. There might be something unnecessary that can be removed.

Preport:
There probably is I didn't look through what the module has to offer and kinda had to botch stuff since I'm pretty new to it.
So it might have unnecessary stuff or might not work cause I couldn't test it.

idinium96 (changes requested):

idinium96:
> There probably is I didn't look through what the module has to offer and kinda had to botch stuff since I'm pretty new to it.
> So it might have unnecessary stuff or might not work cause I couldn't test it.

Yeah, it's okay. Pretty nice that you get to know the flow this quick.

Also, you'll need to also edit the `src/schema/options-json/options.ts` file.
Refer to this PR when adding new options: https://github.com/TF2Autobot/tf2autobot/pull/480/files

Preport:
Updated the file if it looks okay I'll change the other files where I use the options and edit /options-json/options.ts

src/classes/Options.ts
```diff
@@ -391,6 +392,7 @@ export const DEFAULTS = {
         avatarURL: '',
         embedColor: '9171753',
         tradeSummary: {
+            declinedTrade:{enable:false},
```
idinium96, line 395: well, I didn't mean here.
It's here:
https://github.com/TF2Autobot/tf2autobot/blob/f719fefbebd6076c20f4eb92e202cb8a8b4c79ae/src/classes/Options.ts#L117-L118

Preport: Well mb ctrl+f'd tradeSummary and pasted it there without actually looking where I was fixing it rn.

idinium96: Haha understand 😅

Preport: I should probably stop coding late especially like now when is literally 6 am

idinium96:
Looks good, will have a look more and probably test it too later.
Thank you!

Preport:
You're welcome :)

idinium96:
Merging to `development` branch...

PR #659, opened 2021-05-17
The priority settings proposal! (2)

Doomguy46:
note: I am creating this new pull request because I had to change the base branch I was working from, problems here and there, hopefully is a full fix now. Also should fix the formatting in this one.

What does this do? it allows for users to choose how best to prioritize their items in the buy listings by certain settings.
the only setting that's not self explanatory is
typeOfSort

if it is = 1 it is least to greatest price
if it is = 2 it is greatest to least price

I'm sure there are still bugs however hopefully this works well, I think it would be a nice feature to have


idinium96:
Why did you install another `jsonschema` module?
image

Doomguy46:
> Why did you install another `jsonschema` module?
> image

ah, it was coming up with an error. my mistake.

idinium96:
> ah, it was coming up with an error. my mistake.
Then please remove it.

PR #683, opened 2021-05-29
Fix queue getting stuck

Preport:
Following the logs, I was provided. 
This is the only promise that I found hanging after the log message `Sending offer...` and before the log message `Sent offer`. 

So I hope this fixes it 😞.


idinium96 (approved):
Thank you 👍

PR #736, opened 2021-06-22
Highvalue exceptions

idinium96:
image


idinium96:
Looks like it's working:
image
image

PR #754, opened 2021-06-27
fix migration crashing with limited options.json

joekiller:
If a user has a subset of options.json the assumption of keys being present in replaceOldProperties can cause crashes.


idinium96:
> If a user has a subset of options.json the assumption of keys being present in replaceOldProperties can cause crashes.

Oh right, never thought about this. Thank you 👍

idinium96 (approved):

PR #755, opened 2021-06-27
[WIP] Price limit implementation

### LINKED ISSUE #546: Price limits
Currently, the user has no way to specify a price limit for items in their pricelist. They can set a specific price, but prices change all of the time. This makes pricing for items very "all or nothing". You either manually price, or auto-price, and there is no in-between.

I propose a feature to allow the user to set minimum and maximum prices. This could be done using `min.metal`, `max.metal`, `min.keys`, and `max.keys`. If a price update occurs that would be outside of these limits, the price for the item will not update for that "side" of the item's price. This allows the user to set a range on an item's price. I believe that these limits should only work when the item is **not** being banked.

I believe this would result in potentially more profit for users and more trades taking place. This is because the item prices can fluctuate, meaning the user doesn't have to be on the outer end of prices to have an item buy or sell. And because of that fact, they will get more sales.

An example can be seen below:
`!add defindex=5021&intent=sell&min.metal=52&max.metal=52.33` - This would add a Mann Co. Supply Crate Key to the pricelist, and would specify that we want to sell the key for between 52 refined and 52.33 refined. This allows some leeway when selling the key (in the event the price fluctuates by a few scrap metal), while still setting a hard limit on what the user wants to sell the price for.

No alternatives other than specifying an exact price.

**Got any bounty as a reward for anyone fulfills your request?**
N/A

**Additional context**
N/A

idinium96:
Resolves #546

Progress:
- [x] Add new property into pricelist entry
- [ ] Implement logic into methods
- [ ] Commands
- [ ] Test


idinium96:
Maybe next time.

PR #777, opened 2021-07-04
🐛 fix counterOffer's filter undefined error

Preport:


idinium96 (approved):
We will test this together with others 😎

PR #823, opened 2021-07-21
🔥 remove giftedBy tag - temporarily unavailable

idinium96:


idinium96:
@Preport, it turned out that this does not work. There's no error whatsoever, but the tag was not removed, thus I remove this from being implemented.

PR #875, opened 2021-09-02
prepare v4.5.0

idinium96:
## New
- ✨Upgraded Prices.tf (#905) - @Nicklason, @joekiller

## Added
- ✨ an option to show bold characters (#873) 
- ✨🔴 Add resetting item notifications (#879) - @carlosliam

## Update/changes
- 🔄 Update bptf listings (#874, #912) - @arik123
- ⬆ upgrade socket.io-client module for custom pricer #918
- 🔎 always check for possible incorrect listing details (#877)
    - 🔎 check for changes in currencies value too (#878)
- 🔥 remove `graceful-fs` module (6d9115b)
- 🔨 fix not save polldata keys if disable notify (#880)
- 🔎 check banned status only from rep.tf (others as backup) (#887)
    - ✨🔎 an option to check for Marketplace.tf banned status (#881)
- 🚧 add a failsafe on issue #882 (#895)

## Fixes
-  🩹 possible patch for missing existing items (#872)
- 🐛 fix unable to counter offer with only keys (#876) - @Preport
- 🐛 fix bot not playing game after doing other than accepting an offer (#907)
- 🐛 Fixing minor issues with newlines & Rate limiting (#909) - @brysondev


src/lib/bans.ts
```diff
@@ -20,7 +83,7 @@ export function isBptfBanned(steamID: SteamID | string, bptfApiKey: string, user
             {
                 url: 'https://backpack.tf/api/users/info/v1',
                 headers: {
-                    'User-Agent': 'TF2Autobot@' + process.env.BOT_VERSION,
```
joekiller, line 23: Is there a rationale for not identifying the specific version of the software in the user agent?

Just sending the name is sufficient to meet the [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-5.5.3). Providing a version identifier as well could be beneficial because that gives web site operators more discretion as to how they deal with TF2Autobot agents in general. For example if an older version was malfunctioning the site could choose to deal with that specific version versus perhaps blocking the whole agent ID.

idinium96: It was removed because @Nicklason said it's good, but perhaps it's better to keep it.

idinium96: I guess I'll revert the changes made in #888.

Thank you

idinium96:
I am finalizing this, but this still has a problem when creating listings on backpack.tf.
The queue thing in the current @tf2autobot/bptf-listings module is currently still will cause a rate limit if you have thousands of items in your pricelist, and I am not sure where to edit tbh.

joekiller:
Ideally if a rate limit is hit there's some throttle that grows over time. That would probably be wherever the API response is processed. I

idinium96:
> Ideally if a rate limit is hit there's some throttle that grows over time. That would probably be wherever the API response is processed.

Yeah true. I guess I will try to fix it in the next update since Nicklason already shut down pricing on the older prices.tf api

PR #876, opened 2021-09-02
🐛 fix unable to counter offer with only keys

idinium96:
Every time, as far as I am aware.
image

### DIFF
--- a/src/classes/Trades.ts
+++ b/src/classes/Trades.ts
@@ -773,9 +773,15 @@ export default class Trades {
                     return index === -1 ? 0 : Math.pow(3, index);
                 }
 
+                let lockKeys = false;
                 function calculate(sku: PureSKU, side: Dict | number, increaseDifference: boolean, overpay?: boolean) {
                     const value = getPureValue(sku);
                     if (!value) return 0;
+                    if (possibleKeyTrade && sku == '5021;6') {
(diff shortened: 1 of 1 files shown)


idinium96:
Might need to do something with these too:

https://github.com/TF2Autobot/tf2autobot/blob/cfd8bb9c6e78787a8067eb3d0b904c07ce0f70f8/src/classes/Trades.ts#L1052-L1057
https://github.com/TF2Autobot/tf2autobot/blob/cfd8bb9c6e78787a8067eb3d0b904c07ce0f70f8/src/classes/Trades.ts#L1090-L1095

🤔🤔🤔

idinium96:
> The fix I've just committed should also work on trades like this Screenshot_243 Screenshot_244

Woah, thank you! Looks good.

idinium96:
Tested, works with normal items, and yes it still does not work if, for example, I am taking 1 key and only giving 2 ref - the bot will decline the trade.

Preport:
It works just fine on my test bot maybe you don't have enough pure to pay for the key. Or I misunderstood what you're saying.
Screenshot_248
Screenshot_249

idinium96:
Eh weird... I will try again later.

idinium96:
LGTM, merging...

Issue #871, opened 2021-08-31
Resetting Steam notifications for new items


carlosliam:
**Is your feature request related to a problem? Please describe.**
When I log into my bots and they have Steam notifications I always feel pressured to clear them. Often they are just new item notifications.

**Describe the feature you'd like**
I've changed my bots to automatically clear these notifications. I wanted to gauge interest in bringing this change to TF2Autobot, perhaps along with a flag in options to enable it.

**Describe alternatives you've considered**
None.

**Got any bounty as a reward for anyone fulfills your request?**
No.

**Additional context**
Commit showing the changes needed to add this feature (without having a corresponding flag in options): https://github.com/carlosliam/tf2autobot/commit/de9de6f6b898c1be7880e5328279ee3f3f982d96

idinium96:
Thank you, didn't know that thing exists.
Feel free to open a pull request to the `development` branch.

PR #879, opened 2021-09-03
Add resetting item notifications

carlosliam:
Add a listener for the `newItems` event on our `SteamUser` instance. If the number of new items is not zero, the listener `SteamCommunity.resetItemNotifications()` to load our inventory page and reset our item notifications to zero.

Fixes: #871


idinium96 (approved):
Thank you 👍

PR #903, opened 2021-10-19
use new prices.tf system

Nicklason:
Hello

I have been working on a new version of prices.tf. It is an entirely new system that will replace the old one. It includes an improved design and a new pricing algorithm. Most of the code is publicly available on GitHub at https://github.com/prices-tf and docs are available at https://api2.prices.tf/docs.

There are many reasons for me to make the new system. The old system is slow and poorly designed. It is also a major expense to me, both because I provide the service for free, but mostly because of the poor design decisions I have made. I know that many people rely on prices.tf, so instead of shutting it down, I made this new system. It is designed to be scaleable and cheaper to run. I also have my own servers that I am running it on, the old system was not designed to run inside containers and I didn't want to spend time refactoring it.

This is a major change, and the new system will soon replace the old one. Anyone who uses the old APIs should move to the new APIs. The new system works similarly to the old one, but one major part of it has been removed, the snapshots. Prices.tf no longer keeps track of snapshots, instead, it keeps track of listings.

To make it easier for TF2Autobot to move to the new system, I have made this PR. I apologize for the poor code and disrespecting the types, but I just quickly made something that works so people can try out the new system. There might be bugs, but I have done some tests and it seems to work fine.

I don't know when I will shut down the old servers, but I am hoping to do it next month. I will give people some time to move to the new system before I do it though.


Bonfire:
Do these changes remove the option for custom pricers from the TF2Autobot code? If so, is there a way for us to keep such functionality while also using the Prices.tf v2 API? Of course this would require all custom pricers to follow the same structure/endpoints that the v2 API follows.

Bonfire:
Also, are we able to merge this into the `development` branch instead of directly to `master`?

Nicklason:
Closing in favor of https://github.com/TF2Autobot/tf2autobot/pull/905

PR #1000, opened 2022-02-11
[fix] Force stop when docker run is enabled

### LINKED ISSUE #573: Do not use PM2 with docker
I see that we have a Dockerfile in the repo now, so I raise this request.

Could we please bypass pm2 checks upon bot restarts? Like, I'd like it to just terminate itself if I'm running it in a container (we could use env option to detect this, for example).

I suppose there's really no point in running PM2 inside a single process container.

rennokki:
Closes #573

So this is an old issue: bots cannot be restarted via steam command/message if it runs in docker with PM2.

If my logic is not flawed, this should fix the issues by checking if there is a `DOCKER` variable and just terminate the process while doing its cleanup.

The only caveat here is that when running the container, it should contain `--restart=always` and `-e DOCKER=1`

```bash
$ docker run \
    -v $(pwd)/tf2autobot_data/123456:/app/files/123456 \
    -v $(pwd)/tf2autobot_data/ecosystem.json:/app/ecosystem.json \
    --restart=always \
    -e DOCKER=1 \
    tf2autobot/tf2autobot:latest-14.16.0-alpine
```

The change is reverse compatible due to the `DOCKER` variable check.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -35,10 +35,16 @@ const paths = genPaths(options.steamAccountName);
 import log, { init } from './lib/logger';
 init(paths, options);
 
-if (process.env.pm_id === undefined) {
+if (process.env.pm_id === undefined && process.env.DOCKER === undefined) {
     log.warn(
         "You are not running the bot with PM2! If the bot crashes it won't start again." +
             ' Get a VPS and run your bot with PM2: https://github.com/TF2Autobot/tf2autobot/wiki/Getting-a-VPS'
+    );
+}
+
+if (process.env.DOCKER !== undefined) {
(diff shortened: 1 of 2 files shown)


idinium96 (approved):
Looks fine to me. Thank you very much 👍

rennokki:
The merge won't create issues unless `DOCKER=1` is specified in the env variables. So I quick merged it.

Testing can be done with `tf2autobot/tf2autobot:5c9572b88bda0ddccd084d82c63a45efbc221b76-16.13.0-alpine` in case you have Docker & a demo config file in handy.

PR #1025, opened 2022-02-16
Use the new backpack.tf API endpoints

arik123:
Changes to accomodate the API migration.


arik123:
Addresses #1025, needs  [@tf2autobot/bptf-listings pull #1](https://github.com/TF2Autobot/node-bptf-listings/pull/1)

idinium96:
Looks like the `deleteAllListings` thing is having some issue... I am not sure tbh. Didn't touch that one on the @tf2autobot/bptf-listings repo
image

idinium96:
> Looks like the `deleteAllListings` thing is having some issue... I am not sure tbh. Didn't touch that one on the @tf2autobot/bptf-listings repo image

This one, fixed, but...
image

Looks like the old mass delete endpoint is no longer usable.

idinium96:
Unable to update "quantity" once the buy order is created (only restart, or delete then re-create the listings will update it)
image
image

idinium96:
Please note that you can only create one buy order for a certain item (tested with the latest commit - e4fd6f1).
For example, if you add "B.M.O.C (Paint: Australium)" to the pricelist, the bot will create a buy order for Australium painted B.M.O.C, but then if you try to add unpainted B.M.O.C, or any other painted B.M.O.C, it will never get listed.

idinium96:
Also, I think I will remove the "quantity" key, since it's not working for updating that value.

arik123:
I think we should try and ask fisk to fix the issue with quantity not updating

arik123:
Fisk awnsered in next-feedback in backpack.tf discord, that he intends to move quantity to the root document, but it is not implemented in such a way yet ... so I guess we can remove it for now

idinium96:
> Fisk awnsered in next-feedback in backpack.tf discord, that he intends to move quantity to the root document, but it is not implemented in such a way yet ... so I guess we can remove it for now

Noted with thanks. 👍❤

PR #1048, opened 2022-03-21
Rate limit on remove friend fix

brysondev:
Hopefully 4 seconds will be enough 👍 

image


idinium96:
Sorry for the late response, and thanks for the PR.
Yeah looks fine, but I think 5 seconds is much nicer number :3

idinium96 (approved):
Thank you 👍

PR #1049, opened 2022-03-24
🗑️ Add opt-in deletion of seasonal junk untradables

carlosliam:
Only deletes these untradable items:
Noise Maker - TF Birthday
Party Hat
Spirit of Giving
Soul Gargoyle


idinium96:
Sorry for the late response, and thank you for the PR.
Nice feature, I will have a look soon.

idinium96:
Also, since you're adding a new option, please add that new option key here:
https://github.com/TF2Autobot/tf2autobot/blob/d5138bfb92f653d93dd881f5e45430af59cbee3b/src/schemas/options-json/options.ts#L433
It should be something like this;
```ts
                deleteUntradableJunk: {
                    $ref: '#/definitions/only-enable'
                }
```

and then add a new element in the `required` array:
https://github.com/TF2Autobot/tf2autobot/blob/d5138bfb92f653d93dd881f5e45430af59cbee3b/src/schemas/options-json/options.ts#L447

carlosliam:
Sorry I took so long to get back to this. I've updated the options schema definition file with that information.

Issue #92, opened 2020-10-17
Bot Stopped working after windows update


DirtyDrep:
**Describe the bug**

**To Reproduce**

1. Updated windows

**Expected behavior**

**Screenshots and logs**
  Log files are located at `~/tf2autobot/logs`. There are three log files:
 0|tf2autobot  | 2020-10-16 21:32:58 info: Starting bot...
0|tf2autobot  | 2020-10-16 21:32:58 debug: Calling onRun
0|tf2autobot  | SyntaxError: Unexpected token  in JSON at position 0
0|tf2autobot  |     at JSON.parse (<anonymous>)
0|tf2autobot  |     at C:\Users\optim\Desktop\tf2autobot\dist\lib\files.js:29:31
0|tf2autobot  |     at C:\Users\optim\Desktop\tf2autobot\node_modules\graceful-fs\graceful-fs.js:123:16
0|tf2autobot  |     at FSReqCallback.readFileAfterClose [as oncomplete] (internal/fs/read_file_context.js:63:3)
0|tf2autobot  | 2020-10-16 21:32:58 error: tf2autobot failed to start properly, this is most likely a temporary error. See the log:
0|tf2autobot  | package.version: 1.5.3; node: v12.18.4 win32 x64}
0|tf2autobot  | Stack trace:
0|tf2autobot  | SyntaxError: Unexpected token  in JSON at position 0
0|tf2autobot  |     at JSON.parse (<anonymous>)
0|tf2autobot  |     at C:\Users\optim\Desktop\tf2autobot\dist\lib\files.js:29:31
0|tf2autobot  |     at C:\Users\optim\Desktop\tf2autobot\node_modules\graceful-fs\graceful-fs.js:123:16
0|tf2autobot  |     at FSReqCallback.readFileAfterClose [as oncomplete] (internal/fs/read_file_context.js:63:3)
0|tf2autobot  | 2020-10-16 21:32:58 debug: Shutdown has been initialized, stopping... {"err":{}}
0|tf2autobot  | 2020-10-16 21:32:58 debug: Handler finished cleaning up
0|tf2autobot  | 2020-10-16 21:32:58 debug: Waiting for files to be saved
0|tf2autobot  | 2020-10-16 21:32:58 debug: Done waiting for files
0|tf2autobot  | 2020-10-16 21:32:58 warn: Exiting...
PM2           | App [tf2autobot:0] exited with code [1] via signal [SIGINT]
[...]

PR #1070, opened 2022-04-18
Atomic writes

Nicklason:
This will prevent files from becoming "corrupt". Instead of overwriting files, a new file is made and is then renamed which results in the file being overwritten. If something goes wrong then the file that was attempted to be overwritten won't be affected by it.

I also downgraded a dependency because it resulted in a dependency conflict. The invalid dependency came from this pull request https://github.com/TF2Autobot/tf2autobot/pull/996.

Fixes https://github.com/TF2Autobot/tf2autobot/issues/92


idinium96:
Thank you for the PR @Nicklason, I will have a look into this soon.

idinium96:
I wonder why the Nodejs 14.x build test failed.

idinium96:
Hah I see
image

idinium96 (approved):
Thank you. I will assume this'll works just fine.

PR #1071, opened 2022-04-21
🐛 fix requests not being ended

joekiller:
- this should fix network socket leaks

Best practice is to always close the connection. When using requests raw method it is noted that .end is not always used.


idinium96 (approved):
Thank you sir @joekiller

PR #1089, opened 2022-05-03
Probably this will fix user-agent

RobotoLev:
Didn't test this, but it should work.

Initial problem: https://discord.com/channels/664971400678998016/699642379266686997/971112109054255224


idinium96:
Oh, already. Alright thank you

PR #1111, opened 2022-05-22
✅ check for missing OR mismatch listings

idinium96:
The previous automatic refresh listings (or `!refreshlist` command) only check for missing listings and did not check for any listings that have mismatched prices.

This should also check for that and should update if needed.


idinium96:
image

Issue #767, opened 2021-06-30
Add individual item id pricing


joekiller:
I'd like to add a feature where you can price a specific item by its current item id.

## Inital goal is: 

Sell listing for item id.

## Stretch goal is:

Buy listing for specific id or specific asset id. Were you might have a dream item you want to buy and use the asset id. Or you have a specific listing for a individual, you can make a buy listing for that id exclusively.

## Details

**Is your feature request related to a problem? Please describe.**
Sometimes I want to specifically price an item. I cannot currently sell an item with the same sku as another as a differentiated item or price.

**Describe the feature you'd like**
Some command syntax to buy and sell by item id. Ie: `!add id=10151297782&intent=sell&sell.keys=1`

**Describe alternatives you've considered**
I asked in chat and it was indicated that this feature was absent.

**Got any bounty as a reward for anyone fulfills your request?**
Nope.

**Additional context**
Rules of the feature:

  1. Feature adds `![add|update|remove] id=<number>&[Listing settings parameters]`. 
  2. Feature adds `![buy|sell] id=<number>&[buy/sell parameters]`
  3. Any specific item pricing will supercede a sku price.

joekiller:
I seems more realistic for now to just do the assetid because the current asset id is readily avaliable whereas the orginial id doesn't come over trades easily so it's fairly difficult to track that consistently.

idinium96:
> I seems more realistic for now to just do the assetid because the current asset id is readily avaliable whereas the orginial id doesn't come over trades easily so it's fairly difficult to track that consistently.

Yes, also the `econItem` also does not include item's original ID, so yes need to use item's asset ID

idinium96:
Also just a reminder, item's assetid will always change when the item is traded, apply paint, apply custom name, apply custom description, and so on.

PR #1114, opened 2022-05-24
Sell by assetid

joekiller:
another go to implement sell by assetid and fix https://github.com/TF2Autobot/tf2autobot/issues/767


joekiller:
the trick now is to get [PricelistManager.addCommand](https://github.com/TF2Autobot/tf2autobot/blob/8b155a2d6c55acf5c7fa08ee0f126207c2d8ae0d/src/classes/Commands/sub-classes/PricelistManager.ts#L34), buy, and such to use out assetid/sku for pricekeys and then it's done!

idinium96:
Do note that this might cause a conflict when the bot tries to create a sell order:
https://github.com/TF2Autobot/tf2autobot/blob/e4d6c36b2944e62eb487f4aa55d0779ee9c71826/src/classes/Listings.ts#L181

Let's say, the bot has 2 items with the same sku, but we set another one to sell using assetid.

Also, maybe good to always set the pricelist created with item assetid to always be intent=1 (sell only).

And do note that if the trade failed, the item assetid might changed - this might be out of scope, but if the sku for that particular item exists in the pricelist, then the bot might sell that item at whatever price we set on that sku, not on that assetid (conflict I guess).

src/classes/MyHandler/MyHandler.ts
```diff
@@ -1173,7 +1173,11 @@ export default class MyHandler extends Handler {
...
             ourItemsHVCount > 0 // Only check if this not empty
                 ? Object.keys(getHighValue.our.items).some(sku => {
-                      return checkExist.getPrice(sku, false) !== null; // Return true if exist in pricelist, enabled or not.
+                      // look at the high value items and for the given sku see if any have a price enabled for the 'high value sku'
+                      const assetidPriceEnabled = items['our'][sku].some(
+                          item => checkExist.getPrice(item.id, false) !== null
+                      );
+                      return checkExist.getPrice(sku, false) !== null || assetidPriceEnabled; // Return true if exist in pricelist, enabled or not.
                   })
                 : null;
```
idinium96, line 1179: So I think the best way is to add something above L1173, for example

```js
let isAssetidPriceEnabled = false
```
Then at L1179, we can add
```js
if (assetidPriceEnabled) isAssetidPriceEnabled = true
````

File: src/classes/MyHandler/MyHandler.ts [hunk 2]
```diff
@@ -1243,20 +1247,50 @@ export default class MyHandler extends Handler {
...
+                            exchange[which].scrap += Currencies.toScrap(match[intentString].metal);
+                            itemPrices[item.id] = {
+                                buy: match.buy,
+                                sell: match.sell
+                            };
+                        }
+                        return !hasAssetPrice;
+                    });
+                }
+            }
```
idinium96, line 1278: And then use that `isAssetidPriceEnabled` together with `'our' === which`

```js
if (isAssetidPriceEnabled && 'our' === which) {
    // ...
}
```

src/classes/Trades.ts
```diff
@@ -956,9 +956,18 @@ export default class Trades {
...
+                                        if ('our' === side) {
+                                            hasMissingPrices = ourItems[sku].every(
+                                                item => prices[item.id] === undefined
+                                            );
+                                        } else {
+                                            hasMissingPrices = true;
+                                        }
+                                        if (hasMissingPrices) {
+                                            return 0;
+                                        }
```
idinium96, line 970: Add here

```js
if (isAssetidPriceEnabled && 'our' === side) {
    // ...
}
```

idinium96 (commented):
Too many loops will slow down the process,

idinium96:
Also, we need to add for `assetid` parameter in `PricelistManagerCommands/getCommand`

https://github.com/TF2Autobot/tf2autobot/blob/03971b0a6932edf0890279c2081054383eeb8ec6/src/classes/Commands/sub-classes/PricelistManager.ts#L1886-L1894

Also at 
- `PricelistManagerCommands/addbulkCommand`
https://github.com/TF2Autobot/tf2autobot/blob/03971b0a6932edf0890279c2081054383eeb8ec6/src/classes/Commands/sub-classes/PricelistManager.ts#L230-L236
- `PricelistManagerCommands/updatebulkCommand`
https://github.com/TF2Autobot/tf2autobot/blob/03971b0a6932edf0890279c2081054383eeb8ec6/src/classes/Commands/sub-classes/PricelistManager.ts#L1182-L1188
- and at `PricelistManagerCommands/removebulkCommand`
https://github.com/TF2Autobot/tf2autobot/blob/03971b0a6932edf0890279c2081054383eeb8ec6/src/classes/Commands/sub-classes/PricelistManager.ts#L1761-L1767

Plus:
- Add an option param to `PricelistManagerCommands/updatebulkCommand` to auto-add with assetid
https://github.com/TF2Autobot/tf2autobot/blob/03971b0a6932edf0890279c2081054383eeb8ec6/src/classes/Commands/sub-classes/PricelistManager.ts#L517-L530

src/classes/Trades.ts
idinium96: or maybe

```js
if (prices[sku] === undefined && isAssetidPriceEnabled && dataDict && !puresWithKeys.includes(sku)) {
    // ...
}
```

joekiller:
I'll try to hit these thanks!

rgalhos:
Hello! I've got some questions:

- What should happen when a SKU gets disabled by the `!update` command or by a high value trade? Should the assets be disabled too?
- Assets should be filtered out of the `!buy [item name]` command, right? Even if there's only one in stock?

joekiller:
The assets will be individually priced so the feature overrides sku pricing
for a specific item.

Given this understanding, for point 1 the assetid price would not be
disabled because the specific item would be priced.

For point 2, the item would be priced individually first and by sku second.
So in the case of one item left that has specific assetid pricing would be
priced at the specific price overriding sku pricing.

I'm not totally sure how we'll tackle !buy item name with assetid pricing
but again, individual pricing should override sku pricing so ideally it
would contextually pick the appropriate item for !buy. We may need to say,
for unclear items we'd tell the buyer that we cannot discern and ask for
sku or assetid

Sent via mobile

src/classes/Trades.ts
idinium96: Oh, this one if from another file... Ignore it.

idinium96:
Will look into this and also test, soon.

joekiller:
I'll be back in action soon as well. The vacation was extended due to some
COVID stuff. Go figure

Sent via mobile

idinium96:
~~v5~~ - Gonna test this soon, and probably make this ready for v4.16.0

rgalhos:
There are some things that need to be addressed, though.
* The bot crashes when it tries to send a counter offer
* I was thinking about locking `min`, `max`, `intent` (and _maybe_ autoprice?) properties for assets
* Check if autoprice works for assets
* Do not allow `!sell [assetid]`

And, of course, lots of tests and code reviewing - I don't want to be introducing vulnerabilities by accident lol

idinium96:
Ohh... That is still a lot of things to do. I see.
Alright, thank you for the info.

~~v5~~ -> v5

src/classes/MyHandler/MyHandler.ts
joekiller: this was culled

src/classes/MyHandler/MyHandler.ts
joekiller: this is all removed

src/classes/Trades.ts
joekiller: all of these inefficient methods were removed 🥳

joekiller:
> Do note that this might cause a conflict when the bot tries to create a sell order:
> 
> https://github.com/TF2Autobot/tf2autobot/blob/e4d6c36b2944e62eb487f4aa55d0779ee9c71826/src/classes/Listings.ts#L181


This is no longer a problem

joekiller:
The only remaining work is to:

1. remove the assetid from the pricelist when it is traded away
2. don't call for a pricecheck for assetid items that are traded away

Note that I do not plan to implement autopricing for assetids as they are already specifically priced.

idinium96:
It's been a week that I've kinda abandon this. I am truly sorry about it. Thank you very much @joekiller for your latest update.
I will try to do the remaining work as soon as possible.

joekiller:
I replied via email but it didn't seem to post so @idinium96 this was my reply:

It's all good buddy. we're all trying to get this feature over the hump and I wasn't waiting for you if that's what you thought. I was just recording my thoughts in this space in case anybody else had some time. This was 4th of July weekend in the US so I was pretty tied up. But I do plan to finish this actively or passively that's just up to universe. Thanks for your help and your time.

idinium96:
I think it's better to replace this with the item assetid.
I will add new `id` property on the pricelist itself.
image

idinium96:
Nice
image

idinium96:
Hmmm need to do something about this...
image

idinium96:
Let's gooo!
image

idinium96:
> 3. Test what happens when selling a specific item by assetid and also selling assets of the same sku side by side.

Looks like it's working.
image
---
image
image
---
image
(Incoming offers):
image

src/classes/Inventory.ts
```diff
@@ -193,7 +194,18 @@ export default class Inventory {
         return nonTradable.concat(tradable).slice(0);
     }
 
-    getAmount(sku: string, includeNonNormalized: boolean, tradableOnly?: boolean): number {
+    getAmount(
```
joekiller, line 197: With this function gaining parameters it maybe worth just making a params object that can handle optionally so that we don't have to make calls like (true, null, null, false)

Doesn't have to be but it occurred to me looking at all the touch points

joekiller:
I have reviewed the commits up to this point and made my only comment. Great progress!!

src/classes/Inventory.ts
idinium96: I've actually done that, but then when I see 90% need that, then I just make it not optional, but I will consider it.

joekiller: The refactor method in webstorm may make the work much more endurable however I'm unsure.

Check this out: https://www.jetbrains.com/help/webstorm/specific-javascript-refactorings.html#ws_js_destructuring_params_for_function

joekiller: Again, not saying it's necessary but if tempted..

idinium96: Thank you for the suggestion. The latest commits should resolve this :)

joekiller:
I'm removing the draft status of this as it's nearly (already?) done

idinium96:
> I'm removing the draft status of this as it's nearly (already?) done

Yeah... Maybe need to add some Discord webhook alert (currently just on failed to automatically remove it).

idinium96:
Something is wrong... Both of these items are not yet in the pricelist, but the bot failed to automatically add them.
image

idinium96:
Withdrew and then send it to my bot again, and it got added.
image

Hmmmm... gonna wait for someone to sell some painted hats to my bot

idinium96:
Oh nvm, I see what's wrong.

joekiller (commented):
noting that I've reviewed up to this point. Nothing sticks out.

I'm am considering seeing if making the following methods with a params object would make the code more readable. It's unclear immediately what the Boolean switches do from the outside. it may read better if they are Booleans attached to parameters.

inventory.getAmountOfgenerics
inventory.getAmount
pricelist.addPrice
pricelist.hasPrice

idinium96:
Yeah, sure thing.

joekiller:
Message seems to work: 
image

Yeah I did a bit of cleaning. It's revertible if it's too much but I think it's an improvement in readability and intention of code.

Also made the pricechecker a set to de-duplicate stuff.  There was some dead code that had de-duplicated prior so I went for the "spirit of intent"

joekiller:
Doh that's the wrong message. Here's a correct one.

image
image

joekiller (commented):
I'm not sure if there is anything else left. I've reviewed up to this point.

idinium96 (approved):
Thank you very much sir @joekiller and @rgalhos!

joekiller:
It's all good buddy. we're all trying to get this feature over the hump and
I wasn't waiting for you if that's what you thought. I was just recording
my thoughts in this space in case anybody else had some time. This was 4th
of July weekend in the US so I was pretty tied up. But I do plan to finish
this actively or passively that's just up to universe. Thanks for your help
and your time.

Sent via mobile

Issue #1137, opened 2022-06-01
Autobot crashing in random intervals


theKuerbis:
**Describe the bug**
Autobot crashes in random intervals, "picks themselves up" afterwards and sometimes continues to function

**To Reproduce**
none, to my knowledge

**Expected behavior**
idk, not crashing maybe lol

**Screenshots and logs**

![image](https://user-images.githubusercontent.com/64287917/171481794-41b21c36-a143-4ded-af21-1cadf22db1b7.png)

[kuerbisbot.error.log](https://github.com/TF2Autobot/tf2autobot/files/8818183/kuerbisbot.error.log)
[kuerbisbot-2022-06-01.log](https://github.com/TF2Autobot/tf2autobot/files/8818202/kuerbisbot-2022-06-01.log)
[kuerbisbot-trade.log](https://github.com/TF2Autobot/tf2autobot/files/8818184/kuerbisbot.trade.log)

**Additional context**
This startet happening after i moved the bot from a vps to my old smartphone. Usually happens once a day maximum twice. Restarts afterwards and continues to work just fine until it happens again.

Version: 
v4.13.0

idinium96:
This should have been fixed on v4.14.0 and future versions.

PR #1141, opened 2022-06-03
🐛 fix websocket crash when trying to reconnect

idinium96:
Solves #1137

Reference: https://github.com/websockets/ws#server-broadcast

image
-->
image


idinium96:
Will this really make any differences? 🤔

PR #1147, opened 2022-06-07
Discord UI for tf2autobot

RobotoLev:
Using of Discord bot, bot will check Discord chat for commands and answer there, like in Steam.


RobotoLev:
Please review package and package-lock, I might accidentally do something bad there. I installed `discord.js@13.6.0` and `@types/node`. On my win machine was building fine.

src/classes/DiscordBot.ts
```diff
@@ -0,0 +1,57 @@
...
+            return; // obey only admin
+        }
+
+        // TODO: don't handle messages until bot is fully up
+
+        const ans = `I see your "${String(message.content)}"! Answered in Steam for now.`;
+        await message.reply(ans);
+
+        const adminID = new SteamID(this.options.steamOfDiscordAdmin);
+        await this.bot.handler.onMessage(adminID, message.content); // TODO: redirect answer to Discord
```
Bonfire, line 51: It may be worthwhile to instead use slash commands, as Discord seems to be against the idea of reading message contents. Slash commands are also much more versatile and has "richer" interactions

src/classes/Options.ts
```diff
@@ -2295,6 +2299,10 @@ export function loadOptions(options?: Options): Options {
         groups: getOption('groups', ['103582791469033930'], jsonParseArray, incomingOptions),
         alerts: getOption('alerts', ['trade'], jsonParseArray, incomingOptions),
 
+        discordApiKey: getOption('discordApiKey', '', String, incomingOptions),
+        discordAdmin: getOption('discordAdmin', '', String, incomingOptions),
+        steamOfDiscordAdmin: getOption('steamOfDiscordAdmin', '', String, incomingOptions),
```
Bonfire, line 2304: Can we remove this option and instead use the Steam IDs from the bot's `admins` option?

src/classes/DiscordBot.ts
RobotoLev: This will require far more work with code. For now bot commands described in `Commands.ts` and in `Help.ts`. Showing it like slash commands in Discord will require either 3rd place where commands specified (to send the full list to Discord) or big rework for something like `Command` class. I consider first of this variants as bad, and second - as long.

For now I have 2 ideas about it. First - just leave `!` as prefix, which will lead to the least inconvenience for the user (because at this moment all commands go with `!` prefix). Second - to make an option to change prefix in order to manage multiple bots with one Discord bot (I'm not sure if Discord allows to connect more than 1 app to Discord bot account, but if yes - it'll be far more comfortably for the user who runs more than one tf2autobot instance).

For now Discord bot doesn't require any approval and doesn't get any restrictions for reading messages if the bot connected to less than 100 (some says 75) servers. I'm sure this is our case.

src/classes/Options.ts
RobotoLev: I still think what to do, if `ADMINS` contains more than one steamID (because we somehow need to connect admin's discord user id with his steamID). That's why I mentioned that `format of .env is subject to change`.

Gonna make MVP first, maybe I'll find an idea by then.

idinium96:
Will look into this and also test, soon.

src/classes/DiscordBot.ts
```diff
@@ -0,0 +1,77 @@
...
+import SteamID from 'steamid';
+
+export default class DiscordBot {
+    readonly client: Client;
+
+    constructor(private options: Options, private bot: Bot) {
+        this.client = new Client({
+            intents: [Intents.FLAGS.GUILDS, Intents.FLAGS.GUILD_MESSAGES, Intents.FLAGS.DIRECT_MESSAGES]
+        });
+        void this.client.login(this.options.discordApiToken);
```
idinium96, line 16: Maybe put the login not in the constructor (in the `start` instead).
If `login` is a Promise, then it should be handled accordingly.

idinium96:
`start` is only called when the bot is starting.
If people want to use this feature, then they need to setup everything correctly, and yes, crash the bot on start if it's not configured correctly.

src/classes/Bot.ts
```diff
@@ -772,6 +777,17 @@ export default class Bot {
...
+                        if (this.options.discordApiToken) {
+                            log.debug(`Initializing Discord bot...`);
+                            this.discordBot = new DiscordBot(this.options, this);
+                            void this.discordBot.start();
+                        } else {
+                            log.debug('Discord api key is not set, ignoring.');
+                        }
+                        /* eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-call */
+                        return callback(null);
+                    },
```
idinium96, line 790: No need to use try-catch on here (Bot.ts).
Just
```ts
                    (callback): void => {
                        if (this.options.discordApiToken) {
                            log.debug(`Initializing Discord bot...`);
                            this.discordBot = new DiscordBot(this.options, this);
                            void this.discordBot.start()
                                .then(() => {
                                    /* eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-call */
                                    return callback(null);
                             })
                               .catch(err => {
                                   /* eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-call */
                                   return callback(err);
                             });
                        } else {
                            log.debug('Discord api key is not set, ignoring.');
                            /* eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-call */
                            return callback(null);
                        }
                    },
```

RobotoLev: Now I see the logic of implementing it as a promise.

But do we really need to crash the bot on startup? Error on login is not always user's mistake, Discord can simply be unavailable. Maybe in this case will be better to send to Steam chat something like "Error while reaching Discord, check logs or use Steam chat"?

idinium96: If we just send that, then they might just restart the bot again and again until it works, unless the bot automatically retries if it's a server-side problem (status code >= 500).

RobotoLev: Then I offer to crash on 401 and continue with Steam chat in other cases. Possible message: "if https://discordstat.us doesn't indicate any problems right now, you can try to restart. If restarting didn't fix the problem - please ask for help".

idinium96: Alright, that looks good. 👍

RobotoLev: In fact, that's not 401, but 4004 (Gateway Close Event Code). But it doesn't really change the logic.

idinium96: I see. Thank you very much.

src/classes/Options.ts
idinium96: I am also confused about what `steamOfDiscordAdmin` is for.

RobotoLev: That's because ADMINS is an array of ids, and when I was implementing this, I didn't know how to handle the case when it contains more than one ID (because bot should know steamID for each discordID in order for trading commands to work).

Is JSON supported for values in .env and ecosystem? If yes, I can suggest reformatting ADMINS to smth like:
```json
[{"steam": "123", "discord": "123" or null}, <...>]
```
Here "steam" is required, "discord" is optional. If you're going to release this on v5, I consider this variant as possible (it's major update, so there's nothing bad in asking user to edit some stuff which is usually not edited).

idinium96: Sorry for the late reply.
Hmmmm this will be somewhat complicated for new people to understand, and I am pretty sure they will struggle with setup later.

Probably limit Discord interaction with non-trading related commands only, i.e. commands that the bot will send offer will not be included, which are `!buy`, `!sell`, `!withdraw`, `!deposit`, `!withdrawMptf`, `!donateBptf`, and `!premium`.

idinium96:
Thank you for this awesome feature implementation, although I don't agree with the latest changes.
See: [My comment](https://github.com/TF2Autobot/tf2autobot/pull/1147#discussion_r896234137)

src/classes/Options.ts
RobotoLev: And how to whitelist admins in Discord then? Don't whitelist at all is bad; whitelist as I had implemented at that moment, but without binding to steamids - it will work, but doesn't look good.

I was going to provide migration tips in wiki and patchnote. If it won't help with understanding - idk what will. Rolling back is possible, but it will roll back good-working functional, which (imo) is essential to let the Discord UI replace Steam chat UI completely.

Also, options.json is completely json - and bot owners somehow manage to work with it. Is reformatting one exact parameter on major update really that bad?

idinium96: Ok proceed. Options.json is already preset, but whatever.

idinium96 (approved):
Thank you very much @RobotoLev!

PR #1169, opened 2022-06-15
✨ Improvement on the `!block` command

idinium96:
An option to add a reason why a user is blocked for future reference (will be shown in the `!blockedList` command).

Usage: `!block <steamid> [reason]`
Example: `!block 7659358192 Trying to exploit`

You can also add the reason for already blocked users with the same command.

Also, note that people that are automatically blocked (during friend requests, receiving an offer, or sending an offer) will be automatically added with `[onFriendRequest]`, `[onReceivedOffer]`, and `[onSendingOffer]`, respectively.


idinium96:
image

PR #1171, opened 2022-06-16
Use native promises

constancebello:
Refactored the code to use native promises and remove a few dependencies.


idinium96:
Thank you. I will have a look on this and solve the conflict once I'm back home.

idinium96:
Works now.
Previously it just never initializes bptf-listings.

idinium96 (changes requested):
Looks like the process is not terminated when there's an error on start, and it will just be stuck there indefinitely.
Please check for missing `catch` or `throw`. Should crash/terminate the process if there's an error while starting.
image

idinium96 (changes requested):
Stopping the bot also does not immediately terminate the process while it's starting, like how it used to be.
image

src/classes/Bot.ts
```diff
@@ -721,329 +717,249 @@ export default class Bot {
...
-                            if (err) {
-                                /* eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-call */
-                                return callback(err);
-                            }
+        let promise = Promise.resolve();
 
-                            data = v;
+        const checkIfStopping = () => {
+            if (this.botManager.isStopping) throw new Error('Bot stopping');
+        };
```
carlosliam, line 735: This function is called in between each promise in the chain and should handle any shutdowns/signals by throwing an error, skipping any further promises added using `.then(func)`. I'm not sure why it's not working.

carlosliam:
I think returning the promise from the `start` function should fix the error handling issue. Just added a commit which does that.

idinium96:
nah.
image

I think should have `catch` here.
https://github.com/TF2Autobot/tf2autobot/blob/aaf13c5500531c65444945a3353e524c7f260ccd/src/app.ts#L144-L152

`.catch(err => throw err));`

idinium96:
The process is still not terminated/crashed when there's an error on start.
image

carlosliam:
Can you place a log statement before line 734 in `src/classes/Bot.ts` to see if the `checkIfStopping` function is being run after each promise in the chain?

idinium96:
Sure thing. Let me try.
Also, if you want to use my extra account reserved for testing, just let me know.

idinium96:
image

The old one should be here that will reject the `start`
https://github.com/TF2Autobot/tf2autobot/blob/f8f7c9a49bd92faad0cd6da9b5b4b1c5c20d3283/src/classes/Bot.ts#L1028-L1031

I am also not sure if using `let promise = Promise.resolve();` is the right way to do or not.

idinium96:
Maybe bring back
https://github.com/TF2Autobot/tf2autobot/blob/f8f7c9a49bd92faad0cd6da9b5b4b1c5c20d3283/src/classes/Bot.ts#L735
https://github.com/TF2Autobot/tf2autobot/blob/f8f7c9a49bd92faad0cd6da9b5b4b1c5c20d3283/src/classes/BotManager.ts#L58

and reject on error, so that it will `catch` the error in `app.js`

carlosliam:
I am unable to reproduce these issues on my machine at the current commit. I tried with node v16.15.1 and node v18.4.0.

idinium96:
If you're unable to reproduce the error, that means Steam is back alive.
Try adding a line that will throw an error anywhere in the promiseChains, like the one I previously did.

constancebello:
I can't reproduce even when manually introducing an error - the code correctly shuts down upon erroring on my machine. See the attached mov file.

[redacted]

idinium96:
hhmmmm weird.

idinium96 (approved):

idinium96:
Merging this into development.
Thank you very much!

PR #1179, opened 2022-06-20
Rotate polldata

Hhanuska:
Rotate polldata every time the file size exceeds 5 MBs

Some possible QoL changes for the future:
- Add an option to the config to customize the maximum size of each polldata file
  - Increasing the maximum size, while already having polldata files saved at a smaller size will cause the bot to start filling up the older, smaller sized files again, which means we will no longer be able to rely on modification date of the file to correctly sort our trades. All trades will have to be sorted in this case for the stats calculations, which will make it slower.
- Remove duplicate entries
  - Active trades and recently processed trades at the time of switching to a new polldata file will appear in both the old and the new file.
  - These duplicates are handled by using the most recent entry when loading and combining all the polldatas, so they won't cause any inaccuracies in our stats.
  - Keep in mind, removing these duplicates could cause the old file to fall back below the maximum size threshold
- Compress the older polldata files
  - Will slow down stats calculations

### DIFF
--- a/src/classes/Commands/sub-classes/Status.ts
+++ b/src/classes/Commands/sub-classes/Status.ts
@@ -8,6 +8,7 @@ import Bot from '../../Bot';
 import CommandParser from '../../CommandParser';
 import { stats, profit, itemStats, testSKU } from '../../../lib/tools/export';
 import { sendStats } from '../../../lib/DiscordWebhook/export';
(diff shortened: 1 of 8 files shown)


idinium96 (approved):
Tested and looks good for now.
screenshot

Thank you very much ❤

PR #1202, opened 2022-07-09
Authenication with custom-pricer using middleware

EliteOneTube:
Token will be sent when establishing a connection with the server instead of emiting an event for authentication. Removed the listener on the unauthorized event. If a client is not authorized the error will come through the event connect_error.


joekiller (approved):
Looks good. We should note that the unauthorized event is being deprecated for the standard `commect_error`

https://socket.io/docs/v4/emit-cheatsheet/

idinium96 (approved):
Thank you @EliteOneTube

PR #1228, opened 2022-07-17
✅ properly handle inventory fetch

idinium96:
Prevent never-ending "The request is a duplicate and the action has already occurred in the past, ignored this time" on fetching inventory when the bot received and processed a lot of offers in a short period.

image


idinium96:
Hhhhmmm nvm. This might create more problems.
Might update inventory internally.

idinium96:
Hmmm might just be fine. Testing.

PR #1252, opened 2022-07-21
Check bans as class

idinium96:
The problem with previous function based bans check is that, this
https://github.com/TF2Autobot/tf2autobot/blob/a3465225bd7973c502cc1b092e098246765f3bd8/src/lib/bans.ts#L7
and
https://github.com/TF2Autobot/tf2autobot/blob/a3465225bd7973c502cc1b092e098246765f3bd8/src/lib/bans.ts#L19

Might be conflicting when we're processing a lot of offer at the same time.
Doing it in class will make everything isolated, but I am not sure if this will cause a memory leak, or will be managed my GC itself.

### DIFF
--- a/src/classes/Bot.ts
+++ b/src/classes/Bot.ts
@@ -223,7 +223,7 @@ export default class Bot {
     private async checkAdminBanned(): Promise<boolean> {
         let banned = false;
         const check = async (steamid: string) => {
-            const v = new Bans(this, this.userID, steamid);
+            const v = new Bans(this, this.userID, steamid, false);
             const result = await v.isBanned();
             banned = banned ? true : result.isBanned;
         };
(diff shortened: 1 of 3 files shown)


idinium96:
The latest commit [35dde4d](https://github.com/TF2Autobot/tf2autobot/pull/1252/commits/35dde4d6ee124cb52ac05054eec049dc29cae9cd) should help with my concern regarding memory leak, right?

RobotoLev:
In my non-professional opinion, making `v` undefined is unnecessary. GC will collect the object after the function was executed, as it will become unreachable.

idinium96:
> In my non-professional opinion, making `v` undefined is unnecessary. GC will collect the object after the function was executed, as it will become unreachable.

I have the same thought too. Thank you for very much.

idinium96:
Looks good, no significant memory leak, but could be because less trades.

idinium96:
Merging this, I will try implement caching on the next PR.

PR #1272, opened 2022-07-31
🔨 Always set intent to sell if using assetid

idinium96:


joekiller:
thanks for this polish 👍

PR #1273, opened 2022-08-01
🐛 fix high value items not checking for asset id price

joekiller:
This ensures that high value items are entered into the getHighValue map with the appropriate price key

### DIFF
--- a/src/classes/Commands/sub-classes/PricelistManager.ts
+++ b/src/classes/Commands/sub-classes/PricelistManager.ts
@@ -2273,14 +2273,15 @@ export default class PricelistManagerCommands {
             const isPremium = this.bot.handler.getBotInfo.premium;
 
             const list = filter.map((entry, i) => {
+                const priceKey = entry.id ? entry.id : entry.sku;
                 const name = entry.name;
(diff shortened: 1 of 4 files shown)


idinium96 (commented):
Thanks for the PR.
I think this might cause an issue here, but if you have tested and it's all good, then I will merge this soon.
https://github.com/TF2Autobot/tf2autobot/blob/fccc83d8fd02f2b535986ae8e71feaadc9114fdf/src/lib/tools/getHighValue.ts#L71

joekiller:
> Thanks for the PR. I think this might cause an issue here, but if you have tested and it's all good, then I will merge this soon.
> 
> https://github.com/TF2Autobot/tf2autobot/blob/fccc83d8fd02f2b535986ae8e71feaadc9114fdf/src/lib/tools/getHighValue.ts#L71

Yeah it'll blow up if one item isn't on the pricelist while another is. lemme fix it up

joekiller:
also fixed !find not working for pricelist entries with and id

image

joekiller:
okay I squashed and rebased onto development

idinium96 (approved):
Thank you very much 👍

PR #1283, opened 2022-08-01
🔨 Resolve prices.tf 401 issue

idinium96:
Might fix the issue where the bot doesn't even bother to get new JWT from prices.tf. Quite ugly tbh.

image

### DIFF
--- a/src/lib/pricer/pricestf/prices-tf-api.ts
+++ b/src/lib/pricer/pricestf/prices-tf-api.ts
@@ -1,5 +1,5 @@
 import axios, { AxiosRequestConfig, Method, AxiosError } from 'axios';
-import filterAxiosError from '@tf2autobot/filter-axios-error';
+import filterAxiosError, { ErrorFiltered } from '@tf2autobot/filter-axios-error';
 import { PricerOptions } from '../../../classes/IPricer';
 import log from '../../logger';
 
@@ -63,8 +63,8 @@ export default class PricesTfApi {
                 ...headers
             });
         } catch (e) {
-            const err = e as AxiosError;
-            if (err.response && err.response.status === 401) {
+            const err = e as ErrorFiltered;
+            if (err?.status === 401) {
                 log.debug('Requesting new token from prices.tf due to 401');
                 await this.setupToken();
                 return this.authedApiRequest(httpMethod, path, params, data, headers);
--- a/src/lib/pricer/pricestf/prices-tf-api.ts
+++ b/src/lib/pricer/pricestf/prices-tf-api.ts
@@ -123,16 +123,12 @@ export default class PricesTfApi {
     }
 
     async setupToken(): Promise<void> {
-        return new Promise((resolve, reject) => {
-            void PricesTfApi.requestAuthAccess()
-                .then(response => {
-                    this.token = response.accessToken;
-                    resolve();
-                })
-                .catch(err => {
-                    reject(err);
-                });
-        });
+        try {
+            const r = await PricesTfApi.requestAuthAccess();
+            this.token = r.accessToken;
+        } catch (e) {
+            log.error(e as Error);
+        }
     }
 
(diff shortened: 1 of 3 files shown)


RobotoLev:
Reviewing it rn

idinium96:
Hhmmmm something is wrong. It does not wait until one job is completed.

image

image

image

src/lib/pricer/pricestf/prices-tf-api.ts
```diff
@@ -53,24 +53,46 @@ export default class PricesTfApi {
         data?: Record<string, any>,
         headers?: Record<string, unknown>
     ): Promise<B> {
-        try {
+        return new Promise((resolve, reject) => {
             if (this.token === '') {
-                await this.setupToken();
+                this.setupToken()
+                    .then(() => {
+                        return PricesTfApi.apiRequest<B>(httpMethod, path, params, data, {
```
idinium96, line 60: Probably because of this... Let me try remove it

RobotoLev:
I had that situtation today, so I'm watching the log in order to make something more clear. I don't think that happens because of using try-catch instead of promise's .catch() thing.

idinium96:
> I had that situtation today, so I'm watching the log in order to make something more clear. I don't think that happens because of using try-catch instead of promise's .catch() thing.

It's okay I've found the problem. But now another problem came.

idinium96:
Looks like this doesn't help either. It doesn't even update my dummy token.
image

RobotoLev:
So as I thought, it just doesn't re-request token for some reason. I'm pretty sure all was fine when I was implementing fix for v4.13.1.

As I see in my Discord channel history, this strange 401 wasn't appearing on v4.13.1-v5.0.0, but appeared again at v5.0.1. I still don't know what might be reason for it, but that's the pure fact.

idinium96:
> So as I thought, it just doesn't re-request token for some reason. I'm pretty sure all was fine when I was implementing fix for v4.13.1.
> 
> As I see in my Discord channel history, this strange 401 wasn't appearing on v4.13.1-v5.0.0, but appeared again at v5.0.1. I still don't know what might be reason for it, but that's the pure fact.

For me, it started appearing on v4.11.0.
I think I am going to revert changes made in https://github.com/TF2Autobot/tf2autobot/pull/1109/commits/3cbf725086f248611e007ece37aa52e59d5f7a98

RobotoLev:
Wait a second. Can `filterAxiosError` change error format???

RobotoLev:
You've probably cut off `response` from `AxiosError`, so catch doesn't re-request token.

idinium96:
> Wait a second. Can `filterAxiosError` change error format???

Ah dang, you're right. That might be the issue. 🤐

RobotoLev:
That's definitely it. There's no `err.response.status` now (there's no `err.response` even), so bot thinks that it's non-standart error and drops it farther.

idinium96:
> That's definitely it. There's no `err.response.status` now (there's no `err.response` even), so bot thinks that it's non-standart error and drops it farther.

Let me try once again, and then again before this.

idinium96:
Yeah... gonna try the old one.
image

idinium96:
Oh god, the old one works just fine
image

RobotoLev:
That's me who was implementing it (check v4.13.1). At that time 401 was handled well, so now I knew the problem is not here.

joekiller:
Yeah I'd revert. Because the function expects a promise returned but nothing is happening but a void call who knows when it'll resolve. It is likely executing out of order. If you want to keep the new form (but using callbacks blah that's what async is for) you'd need to return the promise instead of executing it as a void

idinium96:
> That's me who was implementing it (check v4.13.1). At that time 401 was handled well, so now I knew the problem is not here.

Yeah, that was before the filterAxios thing.

idinium96:
Thank you @RobotoLev!
Feel free to approve this PR.
image

RobotoLev (approved):

PR #1300, opened 2022-08-08
A new feature: ability for user to force the manual review

RobotoLev:
`force-review` or some other phrases in a trade message -> some accepts and declines won't be applied.

Main use case: managing refund requests.


RobotoLev:
I decided to not suppress declining in some cases (not 5/25 uses, giving with intent sell, taking with intent buy). Will appreciate another opinion about these logics.

I still have to do: an option in `options.json` (probably), checking for admin review being enabled (100%), anti-spam (maybe, not sure yet).

src/classes/MyHandler/MyHandler.ts
```diff
@@ -880,70 +880,48 @@ export default class MyHandler extends Handler {
...
-
-                offer.log(
-                    'trade',
-                    `is a gift offer, accepting. Summary:\n${JSON.stringify(
-                        summarize(offer, this.bot, 'summary-accepting', false),
-                        null,
-                        4
-                    )}`
-                );
+        const forcesReview = offerMessage === 'force-review'; // maybe there is a better keyword?
```
idinium96, line 883: I think it's better to make it not be the exact keyword (use `Array.some` on the list of possible keywords).
Possible keywords:
- `refund`
- `review`
- `check`

RobotoLev:
Sadly don't have time to finalize this. If you do, I will try to assist with reviewing and testing.

idinium96:
I guess it works, but it's missing the mention because I disabled the mention if contains INVALID_VALUE.
I wonder if it's necessary to also calculate the items' value, or just do it like a gift or offer from admins (no value).

image

RobotoLev:
I think we do need to calculate the value, for example for refunding with metal (to see which amount of metal we're going to send)

idinium96:
> I think we do need to calculate the value, for example for refunding with metal (to see which amount of metal we're going to send)

Yeah, that make sense. The bot should automatically subtract the item's value from the profit calculation if refunded for example.

idinium96:
I guess I will just make it mention if contain `⬜_REVIEW_FORCED` reason.

RobotoLev:
Oh, I thought you were talking about amount numbers in the webhook message.

Can't say much about profit calculations since don't know how they're implemented.

idinium96 (approved):
Thank you again 👍

Issue #1354, opened 2022-09-14
feature: custom prefix (discord/steam)


Tolfx:
**Is your feature request related to a problem? Please describe.**
Currently `!` is the default prefix, is good and all as a standard, but when it comes to discord you might don't want the default prefix as `!` since there might be other bots which got the same prefix which causes issues.

**Describe the feature you'd like**
Custom prefix for steam & discord on options

**Describe alternatives you've considered**
For discord you could also instead use slash commands to simplify it and just skip commands. Will make the process way better and enjoyable imo, since you don't need to remember the parameters and get a guide too.

PR #1355, opened 2022-09-14
feature: custom prefix for steam & discord

Tolfx:
# Description
Custom prefix for steam and discord configured in `options.json`
* `options.miscSettings.prefixes.steam`
* `options.miscSettings.prefixes.discord`

# Changes
* Updated schema validation for prefixes options
* Updated logic for custom prefixes
* Updated all messages with static fixed prefix to custom prefix

# To do
* Update wiki
* Test cases for safety
  * Tested locally and worked what I could see, but rather new to the system so should test everything to see it works as intended  

# Extra 
Fixes: #1354 

*Wanted to create %prefix% for messages on bptf.*


idinium96:
Thank you for the PR. I will look into this.

RobotoLev:
Interesting to see this. When I was implementing Discord interface from scratch, there were some discussions about differentiating bots by their prefixes, but that time we decided to manage it by differentiating Discord channels bots have access to.

I have some vague ideas about OOP-reworking command interface to make managing commands easier for programmers, as well as moving to slash-commands in Discord. Howewer, your PR might be good for some users for now.

Tolfx:
> Interesting to see this. When I was implementing Discord interface from scratch, there were some discussions about differentiating bots by their prefixes, but that time we decided to manage it by differentiating Discord channels bots have access to.
> 
> I have some vague ideas about OOP-reworking command interface to make managing commands easier for programmers, as well as moving to slash-commands in Discord. Howewer, your PR might be good for some users for now.

I thought about making a whole module for commands and extensions for it but I'm selfish and don't need it. But hopefully it will in the future since the current way of handling commands scares me, would be easier to have them into classes and have a `CommandHandler` containing all the commands.

src/classes/Commands/Commands.ts
```diff
@@ -94,7 +94,8 @@ export default class Commands {
     }
 
     async processMessage(steamID: SteamID, message: string): Promise<void> {
-        const command = CommandParser.getCommand(message.toLowerCase());
+        const customSteamPrefix = this.bot.options.miscSettings?.prefixes?.steam ?? '!';
```
idinium96, line 97: As far as I am aware, the Discord UI also use this `processMessage` method.
Maybe use `this.bot.getPrefix(steamID)` instead of this?

Tolfx: Fixed

Tolfx:
Wouldn't it be better to refactor the whole command system, using if statements to check what command the user is using is too much bloat code, wouldn't it save more time to create a `Command` class with a `CommandHandler` getting our message if it's a command request and find our command in a map/array to which have been cached on bootstrap?

Would be easier to handle issues like this prefix issues because going around finding where `!` is used will be too much repeating process, to avoid **DRY** so to say.

idinium96:
If you're able to do that, sure, but let this PR get merged first, and I'll give you a write permission on the main repository so that you'll be able to create another branch for that.

src/classes/CommandParser.ts
```diff
@@ -3,11 +3,9 @@ import { UnknownDictionaryKnownValues } from '../types/common';
...
 
 export default class CommandParser {
-    static getCommand(message: string): string | null {
-        if (message.startsWith('!')) {
-            const index = message.indexOf(' ');
-
-            return message.substring(1, index === -1 ? undefined : index);
+    static getCommand(message: string, prefix: string): string | null {
+        if (message.startsWith(prefix)) {
+            return message.slice(prefix.length).trim().split(/ +/g).shift()?.toLowerCase();
```
idinium96, line 8: I actually wondering what's that `split(/ +/g)` for. Could you explain it?

Tolfx: If the command goes like this: `!help 1231323` it will split the space between and shift it to only get `help` and not any other arguments for the command. If that explains it.

Tolfx: I've used this on my bot without any issues, so don't think so, could make a test case for it to try?

idinium96: Ah, I see. Forgot that the ` +` is a thing in regex.
I guess I will make it split the first space (no need for a global tag).

Thanks for the explanation.

idinium96 (approved):
Thank you for the PR, and really sorry for the delay.

PR #1348, opened 2022-09-08
Discord improvements

RobotoLev:
A couple of things related to the Discord module of tf2autobot

PR #1381, opened 2022-10-16
Revert "Discord improvements"

idinium96:
Reverts TF2Autobot/tf2autobot#1348

@RobotoLev, looks like there's something that prevents the bot from starting.
I am reverting the changes for now.

image


RobotoLev:
That line probably requires checking for `admin.discordID` being not `null`. Can't test right now, sorry about that.

PR #1396, opened 2022-10-28
Fix for #1348

RobotoLev:
Didn't test it (I'm on phone right now), but it should work. 

The problem is: I committed an old code without testing it properly. Or the last commit in #1348 was conflicting with other ones, can't say for sure right now.


idinium96 (approved):
Yep, this resolves the crashes. Thank you very much.

idinium96:
Btw, is it supposed to be `null`? If not, why most of the time it's `null`?

RobotoLev:
Actually, it doesn't do what it should. Just tested, dm chats became unavailable with this commit. Gonna fix it again.

idinium96:
> Actually, it doesn't do what it should. Just tested, dm chats became unavailable with this commit. Gonna fix it again.

Alrighty. 👌

PR #1397, opened 2022-10-29
Another fix for #1348

RobotoLev:
Found out that I used a wrong method to construct `User` object by its id.

I have tested this and it seems to work normally.


idinium96 (approved):
I see. It's an async function. I've tested it and it works.
Thank you 👍

PR #1413, opened 2022-11-02
➕ Add `!groups` command

RobotoLev:
This command was requested a couple of times, so I decided to implement it.

How to use: as an admin, send `!groups` to get a list of groups in your bot's pricelist.


idinium96 (approved):
👍 Thank you very much!

PR #1421, opened 2022-11-10
More Discord fixes

RobotoLev:
Yet another bunch of fixes related to the Discord interface.

- Fixed crash when Discord bot tries to start, but fails
- Improved validation of admins' ids (to disallow having different steam ids with the same non-null discord id)

Maybe something else will be added, but idk ¯\\\_(ツ)_/¯


idinium96 (approved):
Looks good, thank you very much.

PR #1435, opened 2022-11-28
🌐 Add options for alternate inventory APIs

carlosliam:
Adds an option to use alternate inventory API endpoints when fetching inventories through the Steam Community API. Does not help with using the IEconItems_440 API. Currently SteamSupply and SteamApis are supported.

If this gets merged I can add all the relevant information to the wiki regarding how to set it up.


src/classes/InventoryApis/InventoryApi.ts
```diff
@@ -0,0 +1,180 @@
...
+    // Adapted from node-steamcommunity
+    public getUserInventoryContents(
+        userID: SteamID | string,
+        appID: number,
+        contextID: string,
+        tradableOnly: boolean,
+        callback: (err?: Error, inventory?: EconItem[], currency?: EconItem[], totalInventoryCount?: number) => void
+    ): void {
+        if (!this.isEnabled()) {
+            callback(new Error("This API is disabled in the bot's options"));
```
idinium96, line 44: I wonder whether this error message will be shown to whoever that's communicating with the bot or not.
If not, then it's alright I guess.

src/classes/InventoryGetter.ts
```diff
@@ -0,0 +1,35 @@
...
+        if (steamID.toString() == this.bot.client.steamID.getSteamID64()) {
+            this.bot.manager.getUserInventoryContents(steamID, appid, contextid, tradeableOnly, callback);
+            return;
+        }
+        for (const inventoryApi of this.inventoryApis) {
+            if (inventoryApi.isEnabled() && inventoryApi.getApiKey() != '') {
+                inventoryApi.getUserInventoryContents(steamID, appid, contextid, tradeableOnly, callback);
+                return;
+            }
+        }
```
idinium96, line 32: Does this mean if both are enabled, it will use both APIs to get the data?

src/classes/InventoryApis/InventoryApi.ts
carlosliam: `InventoryGetter` checks whether each inventory API is enabled before using this method on it, so this error should not ever come up in normal bot usage.

src/classes/InventoryGetter.ts
carlosliam: The `return` statement ensures that only the first valid `inventoryApi` gets called.

idinium96: Oh yeah didn't realized it. Thank you.

src/classes/InventoryApis/InventoryApi.ts
idinium96: I see. Alright thank you very much.

idinium96 (approved):
So far, the bot works fine with the default API, and I haven't tested the API for `steamapis` and `steamsupply`, but based on my evaluation of the code review, it should work as intended.

Thank you very much for this implementation!

PR #1459, opened 2023-01-05
Add an option to decline CRIME_ATTEMPT instead of counter the offer

idinium96:
- Revert #1455 and add an option to decline instead.


brysondev (approved):
Logic looks good to me!

brysondev (approved):
Simple, effective, clean. Looks good to me!

brysondev (approved):
Okay, this is actually the most clean way of implementing this code, now we look good!

I swear 😤

PR #1484, opened 2023-01-26
New option to auto-enable halt mode on start

RobotoLev:
Pretty simple thing to be sure that your bot won't trade with customers

### DIFF
--- a/src/classes/Bot.ts
+++ b/src/classes/Bot.ts
@@ -1185,11 +1185,6 @@ export default class Bot {
                         return resolve();
                     }
 
-                    // TODO: was it necessary?
-                    // if (this.options.discordBotToken && this.discordBot && !this.halted) {
-                    //     this.discordBot?.setPresence('online');
-                    // }
-
                     this.manager.pollInterval = 5 * 1000;
                     this.setReady = true;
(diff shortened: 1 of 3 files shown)


RobotoLev:
There is a warning on startup, which is caused by calling `halt()` until the booting is complete (removing all listings fails at that moment). It doesn't affect anything, so I'm not sure if it has to be removed.

Also there's a copypasta leftover in the code. I don't know is it necessary, so commented it and marked with TODO. Please uncomment or remove it before merging.

src/classes/Bot.ts
```diff
@@ -1182,9 +1185,10 @@ export default class Bot {
                         return resolve();
                     }
 
-                    if (this.options.discordBotToken && this.discordBot) {
-                        this.discordBot?.setPresence('online');
-                    }
+                    // TODO: was it necessary?
+                    // if (this.options.discordBotToken && this.discordBot && !this.halted) {
+                    //     this.discordBot?.setPresence('online');
+                    // }
```
idinium96, line 1191: Yes, I just realize it's duplicated. Thanks for it.

idinium96 (approved):
Thank you for the PR!

PR #1518, opened 2023-04-01
fix: 🐛 "Voices From Below" renamed to "Voices from Below"

joekiller:
Per the [March 30th patch](https://wiki.teamfortress.com/wiki/March_30,_2023_Patch) the inventory now returns Voices from Below

Also added DeprecatedJsonOptions to allow migration of options in predicable way.

This fixes starting you bot and getting the following error because Voices From Below returns an "undefined" key with no high value type.

>  {"level":"error","message":"TF2Autobot failed to start properly, this is most likely a temporary error. See the log:\r\npackage.version: 5.6.0; node: v16.13.2 win32 x64}\r\nStack trace:\r\nTypeError: Cannot read properties of undefined (reading 'replace')\n    at c:\\tf2autobot\\dist\\classes\\Listings.js:463:61\n    at Array.forEach (<anonymous>)\n    at Listings.getDetails (c:\\tf2autobot\\dist\\classes\\Listings.js:431:37)\n    at Listings.checkByPriceKey (c:\\tf2autobot\\dist\\classes\\Listings.js:228:35)\n    at Immediate.<anonymous> (c:\\tf2autobot\\dist\\classes\\Listings.js:325:30)\n    at processImmediate (node:internal/timers:464:21)\r\nBot has been up for 31 seconds.","timestamp":"2023-04-01T14:54:38.907Z"}
{"err":"uncaughtException","level":"debug","message":"Shutdown has been initialized, stopping...","timestamp":"2023-04-01T14:54:38.908Z"}

### DIFF
--- a/.example/options.json
+++ b/.example/options.json
@@ -787,7 +787,7 @@
             "Chromatic Corruption": "CC 🪀",
             "Spectral Spectrum": "Spec 🔵🔴",
             "Sinister Staining": "Sin 🍈",
-            "Voices From Below": "VFB 🗣️",
+            "Voices from Below": "VFB 🗣️",
             "Team Spirit Footprints": "TS-FP 🔵🔴",
             "Gangreen Footprints": "GG-FP 🟡",
             "Corpse Gray Footprints": "CG-FP 👽",
--- a/.example/options.json
+++ b/.example/options.json
@@ -787,7 +787,7 @@
             "Chromatic Corruption": "CC 🪀",
             "Spectral Spectrum": "Spec 🔵🔴",
             "Sinister Staining": "Sin 🍈",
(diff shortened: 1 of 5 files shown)


idinium96 (approved):
Thank you very much for quick fix!

joekiller:
@idinium96 why revert the options example? The options loaded will always modify the options to be the lowercase version so that isn't really a true opinion

idinium96:
> @idinium96 why revert the options example? The options loaded will always modify the options to be the lowercase version so that isn't really a true opinion

The reason I reverted it is that I usually do it once a new version is released (direct commit to the `master` branch), so that people can see what has been changed on that file.

joekiller:
@idinium96 thanks for the clear explanation ❤️

Issue #272, opened 2021-01-13
Enhance key conversion rate


sumet123:
**Is your feature request related to a problem? Please describe.**

The current implementation use key selling price for conversion for calculating the price. Consider the following situation:
_The market (other bots on classified) is buying a key for 51 refs and selling for 52 refs - selling for more than 52 refs is almost impossible. Somehow, our autoprice price the key to buying for 51 refs and selling for 53 refs. Someone could take 1-ref item and 52 refs from the bot by paying a key - meaning the bot paying ~53 refs for a key. However, this problem shouldn't last that long because the autoprice will update the price to meet the market._

**Describe the feature you'd like**

Use different key conversion rate base on 2 scenarios:
- Use key **selling price** for conversion when the partner is **taking keys** from the bot
- Use key **buying price** for conversion when the partner is **giving keys** to the bot

This implementation is more logical, all other items use this logic for calculating the price.

idinium96:
![image](https://user-images.githubusercontent.com/47635037/104728795-68bc6f00-5772-11eb-93f8-8356195127dc.png)

Because of this, I've canceled this implementation.

sumet123:
> ![image](https://user-images.githubusercontent.com/47635037/104728795-68bc6f00-5772-11eb-93f8-8356195127dc.png)
> 
> Because of this, I've canceled this implementation.

That happens because the bot will convert everything (including keys) to scrap for calculating the value. All incoming offer buying the items worth more than 1 key (`sell.keys > 0`) will be declined because of invalid value (unless the partner paying with metals only)

The possible fix is to not convert the key to scrap (both actual key and the key from `buy.keys` / `sell.keys` in the item price) when calculating the value. Then using { key: number, value: number (value in scrap) } for calculating and comparing instead.

My solution takes a lot of rework on the complicated parts of the bot so it is not worth implementing just for this issue. There might be better solutions but I have no other idea for now.

idinium96:
yes and I against your solution.

joekiller:
@idinium96 please correct me if I'm wrong.

The intention is:

The buying rate is only applied if the key being specially bought by the bot.

The selling rate is applied if the key is being offer in exchange for an item that is for sale. 

I think this can be closed.

Issue #1506, opened 2023-02-28
Protect Against Abusive Currency Exchange During Overpay


joekiller:
**Is your feature request related to a problem? Please describe.**
People can offer a key for a 0.05 priced item and the bot will dump all the metal it has for the key even if it isn't selling metal for keys.

**Describe the feature you'd like**
I had a few ideas:
 a. Create an option to deny trades where a key is not in the sale price of the item.
 b. Create an option to limit the overpay of currency to a certain percentage of the sale price, say up to ~10% overpay of the value of the item, to help make change.

PR #1520, opened 2023-04-02
🔨🔥 key rate => use sell price if our side, buy price if their side

idinium96:
A possible solution for issue #272.

Note: This might affect the profit calculations at the moment.


src/classes/MyHandler/MyHandler.ts
```diff
@@ -678,6 +680,12 @@ export default class MyHandler extends Handler {
                     } else if (sku === '5021;6') {
                         exchange.contains.keys = true;
                         exchange[which].contains.keys = true;
+                        if (which === 'our') {
+                            keyOurSide = true;
+                        } else if (which === 'their' && keyOurSide === true) {
```
RobotoLev, line 685: Are you sure that `else if` should be here and not just `if`?

idinium96: How can I determine that there are keys on both sides if I didn't do like that?

joekiller:
I dunno if this addresses #1506 but yeah seems to fix the intention of only paying a "buying" rate when truly buying keys.

idinium96:
> I dunno if this addresses #1506 but yeah seems to fix the intention of only paying a "buying" rate when truly buying keys.

Oh yeah, I guess I tagged the wrong issue. Oppss :3

src/classes/MyHandler/MyHandler.ts
RobotoLev: Right, now I've managed to catch the logic. Changing `else if` to `if` wouldn't change anything, since `which` can't be `their` and `our` at the same time. But I still kinda dislike the logic. Because if for some reason we change the order of processing sides (so that "their" is processed before "our"), this feature will break.

idinium96: Yea, anyway gonna keep it for now.

joekiller:
This should fix https://github.com/TF2Autobot/tf2autobot/issues/272

idinium96:
I guess I will merge this now, and I will think about `UserCart.ts` later.

PR #1558, opened 2023-06-04
Use alternative delete if needed

joekiller:
This works around the bot getting into an infinite loop/flaming startup death if the bp mass delete listing endpoint isn't working.

It patches in the currently deleted ListingManager.removeListings function. Once [this pr to add removeListings back](https://github.com/TF2Autobot/node-bptf-listings/pull/9) is accepted, the patching can be removed and just the call can be used.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
             "version": "5.7.0",
             "license": "MIT",
             "dependencies": {
-                "@tf2autobot/bptf-listings": "^5.6.11",
+                "@tf2autobot/bptf-listings": "^5.7.1",
                 "@tf2autobot/bptf-login": "^2.3.5",
                 "@tf2autobot/filter-axios-error": "^1.5.1",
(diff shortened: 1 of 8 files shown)


arik123 (approved):

idinium96:
Thank you very much for the PR sir @joekiller! I will look into this very soon. Sorry for the delay.

idinium96:
[This PR](https://github.com/TF2Autobot/node-bptf-listings/pull/9) on @tf2autobot/bptf-listings has been merged and the `removeListings` should be available on the latest version. To prevent any issues, I am appointing sir @joekiller to update this PR accordingly (please do only in your free time, no rush).

joekiller:
Thanks @idinium96 I'll likely hit it this weekend

joekiller:
@idinium96 I have updated the dependency and removed the patching.

also added some abort and retry once behavior with github user content connections sometimes hanging at startup.

PR #1560, opened 2023-06-04
Max shealy patch 1

MaxShealy:


template.ecosystem.json
```diff
@@ -1,3 +1,6 @@
+// This is a template for your ecosystem.json file. Refer to the wiki to learn how to set it up. 
+// (Note: you must delete "template." for it to work)
+
```
idinium96, line 3: I appreciate the effort, but you can't add comments on a `.json` file. People that don't know will have a problem launching the bot if they didn't delete this before it.

MaxurDev: You're right about that, that was my mistake.

PR #1602, opened 2023-07-17
[WIP] Internal backpack

idinium96:
Still work in progress.
Instead of fetching the bot backpack every time after a successful trade, we manage it internally.


idinium96:
Maybe next time.

Issue #1682, opened 2023-10-27
Sell commands do not work with generic unusuals


purplebarber:
Generic unusual sell commands do not work from the times I've tested.

Steps to reproduce the behaviour:

1. Have a bot buying generic unusual, and have an unusual in your inventory
2. Use !sell (unusual)
3. Bot fails to construct an offer (error: you do not have x in your inventory)

**Expected behavior**
Bot constructs offer with first unusual that falls under generic

**Screenshots and logs**

<img width="621" alt="image" src="https://github.com/TF2Autobot/tf2autobot/assets/87976511/09733569-0a87-4955-88dc-15d9bb1f9c9a">

Version: 5.8.4

PR #1683, opened 2023-10-27
fix sell commands not working for generic unusuals

purplebarber:
fixes commands like !sell (generic unusual).

add findByPartialSku() for Inventory.ts which is called when an unusual object has no effect. Is called when searching for generic unusuals in other player's inventory.

fixes: https://github.com/TF2Autobot/tf2autobot/issues/1682

### DIFF
--- a/src/classes/Inventory.ts
+++ b/src/classes/Inventory.ts
@@ -196,7 +196,7 @@ export default class Inventory {
         return nonTradable.concat(tradable).slice(0);
     }
 
-    findByPartialSku(partialSku: string, tradableOnly = true, elevatedStrange = false): string[] {
+    findByPartialSku(partialSku: string, elevatedStrange = false): string[] {
         const matchingSkus: string[] = [];
 
         if (elevatedStrange) {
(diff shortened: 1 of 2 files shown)


src/classes/Inventory.ts
```diff
@@ -195,6 +196,44 @@ export default class Inventory {
...
+                }
+            }
+
+            if (!tradableOnly) {
+                for (const sku in this.nonTradable) {
+                    if (sku.startsWith(partialSku) && sku.includes(';strange')) {
+                        matchingSkus.push(...this.nonTradable[sku].map(item => item?.id));
+                    }
+                }
+            }
```
idinium96, line 217: I think this is not necessary, as it's impossible to add items that are not tradeable.

src/classes/Inventory.ts
```diff
@@ -195,6 +196,44 @@ export default class Inventory {
...
+
+            if (!tradableOnly) {
+                for (const sku in this.nonTradable) {
+                    if (sku.startsWith(partialSku)) {
+                        matchingSkus.push(...this.nonTradable[sku].map(item => item?.id));
+                    }
+                }
+            }
```
idinium96, line 232: Same with above.

idinium96 (changes requested):
Thanks for the PR!
Have you tested this?

purplebarber:
Yups tested it, and attached a screenshot of it working. Other unusual sell orders work fine too. I'll push commits to change the tradeable checks.

image

src/classes/Inventory.ts
purplebarber: Removed check

src/classes/Inventory.ts
purplebarber: Removed check

idinium96 (approved):
Thank you very much!

PR #1708, opened 2024-03-31
Bump Steam related libraries

idinium96:


idinium96:
Something went wrong here, the bot does not detect incoming trade offer, and also does not process accepted trade offer (created via chat)

PR #1746, opened 2024-05-01
feat: api key in env

Hhanuska:
Api key can no longer be obtained automatically.
https://github.com/DoctorMcKay/node-steamcommunity/issues/342

We can set it manually in ENV


idinium96 (approved):
Thank you very much!

PR #1778, opened 2024-07-18
Include package-lock.json in Docker build

juniorISO69960:
Fixes an issue where the latest version of typescript cannot compile TF2Autobot by removing the package-lock.json from the .dockerignore so it is included in the build process.


idinium96 (approved):
Thank you very much for your help!

PR #1780, opened 2024-07-24
!rate properly displays a rate

Konaya245:
Previously only reiterated key sell rate, now shows clearly states buy/sell rate.


idinium96:
Thank you very much for this! Everything looks fine and I will merge with the next update.

PR #1781, opened 2024-07-26
Add an option for denying trades with items on both sides of the trade.

juniorISO69960:
Adds an option for controlling whether or not a trade offer will be accepted / processed if both our and their side of a trade offer contain items other refined metal and keys.

Setting: miscSettings.itemsOnBothSides

Also fixes a typo in the sendTradeDeclined preventing the webhook from properly displaying the bot's avatar if not manually set in the options.

🫡


idinium96 (approved):
Thank you very much for this feature. Much appreciated and sorry for the long wait.

PR #1795, opened 2025-05-21
Disable counter offers

arik123:


idinium96:
Thank you

Nicklason:
There is already an object for the settings so it would have been better to just disable it in that object. https://github.com/TF2Autobot/tf2autobot/blob/2931e4662fb722203a7d58507efbf6a039dd4c50/src/classes/Options.ts#L2489 https://github.com/TF2Autobot/tf2autobot/blob/2931e4662fb722203a7d58507efbf6a039dd4c50/src/classes/Options.ts#L30-L34

PR #1797, opened 2025-05-22
fix: don't send offers with invalid items

Nicklason:


idinium96 (commented):
Looking back at the counterOffer method, I wonder if the `BigInt(item.contextid) <= 0` is necessary. Probably only need to check for the `item.assetid`?

https://github.com/TF2Autobot/tf2autobot/blob/017638c7bf4b7b27385c3541fe81c12a6e06ff04/src/classes/Trades.ts#L790-L796

https://github.com/TF2Autobot/tf2autobot/blob/017638c7bf4b7b27385c3541fe81c12a6e06ff04/src/classes/Trades.ts#L1009-L1013

Nicklason:
> Looking back at the counterOffer method, I wonder if the `BigInt(item.contextid) <= 0` is necessary. Probably only need to check for the `item.assetid`?
> 
> https://github.com/TF2Autobot/tf2autobot/blob/017638c7bf4b7b27385c3541fe81c12a6e06ff04/src/classes/Trades.ts#L790-L796
> 
> https://github.com/TF2Autobot/tf2autobot/blob/017638c7bf4b7b27385c3541fe81c12a6e06ff04/src/classes/Trades.ts#L1009-L1013

It is 100% required because the item in the offer is copied over from offer.counter().

PR #1851, opened 2025-09-15
Added pricedb.io as a pricer replacement for prices.tf 

OliverPerring:
Added pricedb.io as a pricer replacement for prices.tf

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,15 +1,15 @@
 {
-    "name": "tf2autobot",
-    "version": "5.13.3",
+    "name": "tf2autobot-pricedb-io",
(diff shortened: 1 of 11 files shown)


package-lock.json
Preport: This entire diff shouldn't be in this diff

package.json
Preport: This entire diff shouldn't be in this diff too

OliverPerring:
> You are changing the package name, version, core dependencies and also modified most of the links to your own fork. Did you vibe code this? 90% of the changes are just useless if not malicious.

Meant to close before changing stuff on the fork my bad. Ignore it for now

PR #1937, opened 2026-07-10
🔨 Bypass getTradeOfferURL if already cached/created

idinium96:
Prevent Steam 429 error during starting the bot.

Create a file named `tradeOfferUrl.txt` inside the `.../tf2autobot/files/<botSteamUsername>` directory/folder, then copy and paste your bot's Steam Trade Offer URL, and then start the bot.

image


idinium96:
image

Hmmm I wonder how long I need to wait for this 429 from Steam.

PR #1943, opened 2026-07-10
retry mobile confirmation up to 5 times with backoff

hashdashme:
- Retry `acceptConfirmation` up to 5 times with 5s/10s/15s/20s delays when Steam is flaky
- If the error is `Could not find confirmation for object`, treat it as success and stop retrying, the
  confirmation is already accepted or no longer exists, so further attempts won't help
- Keeps existing debug logs and add a warn log on each retry


src/classes/Trades.ts
idinium96: I think these will need to be removed if we're doing the mobile confirmation retry on `acceptConfirmation`, otherwise it might cause conflicts or the actual rate limit.

Still reviewing on other things such as this (errors ignored for I don't know why for chat commands, probably because the offer will be cancelled automatically in 1.5 minutes if not confirmed):
https://github.com/TF2Autobot/tf2autobot/blob/76d21876f4c220adfb5cfaaa569a8d85443788e9/src/classes/Carts/CartQueue.ts#L266-L268

Feel free to do something if it's necessary and thank you very much for the pull request.

idinium96 (approved):
Tested and working.

image
image
image

Steam chat:
image

PR #1944, opened 2026-07-10
Fix Steam 429 on Escrow checks

idinium96:
Use WebAPI escrow checks before HTML fallback

Cherry-pick commit from @Hexenum's fork of @OliverPerring's (PriceDB) fork.


idinium96:
Tested and working.

